PR #1, opened 2024-11-12
Fission lib, IEx WASM

mat-hek:
Adds two new sub-projects:

- FissionLib, which contains all the Erlang and Elixir libs that ship with AtomVM and uses them as patches to the original stdlibs (taken from the OS). 
   The patching works by decompiling beams to Core Erlang, merging respective modules' implementations, and compiling them back. Initially, I tried renaming and keeping both implementations and creating a proxy module that would call either the original implementation or the patch. It didn't work for two reasons: the original implementation sometimes has local calls to itself, and NIFs can be loaded into a single module with a name specified in the NIF.
- IExWASM - for now, it's a textarea executing Erlang code using `erl_eval`. Based on the FissionLib.

For details, see the readmes of FissionLib and IExWASM.

The first 3 commits contain files copied from AtomVM, the rest is the actual code.


iex_wasm/static/.gitignore
```diff
@@ -0,0 +1 @@
+/iex_wasm.avm
```
jgonet: Newline missing.

fission_lib/libs/libdupatrace/dupa_trace.erl
```diff
@@ -0,0 +1,135 @@
+-module(dupa_trace).
```
jgonet: Not particularly fond of name, those things like to leak.

PR #6, opened 2024-11-25
Make prim_eval work

mat-hek:
- Copy prim_eval.S to FissionLib - no idea how to make it work otherwise, but didn't dig
- Mock erlang:bump_reductions - see https://linear.app/swmansion/project/erlangbump-reductions-missing-02f6ff09cac1/overview
- Mock erl_anno:set_file - see https://linear.app/swmansion/project/erl-annoset-file2-sometimes-crashes-in-erl-eval-76e146323242/overview
- Add tests 🤯

### DIFF
--- a/fission_lib/config/config.exs
+++ b/fission_lib/config/config.exs
@@ -1,3 +1,9 @@
 import Config
 
 config :fission_lib, add_tracing: true
+
+if Mix.env() == :test do
+  # Put the following configuration in config.secret.exs:
+  # config :fission_lib, atomvm_path: ...
+  import_config "config.secret.exs"
+end
--- a/fission_lib/lib/fission_lib.ex
+++ b/fission_lib/lib/fission_lib.ex
@@ -33,7 +33,7 @@ defmodule FissionLib do
 
     :packbeam_api.create(
       ~c"#{options.output_path}",
-      Enum.map([options.fission_lib_path] ++ options.artifacts, &String.to_charlist/1),
+      Enum.map([options.fission_lib_path | options.artifacts], &String.to_charlist/1),
       %{start_module: options.start_module}
     )
     |> case do
--- a/fission_lib/config/.gitignore
+++ b/fission_lib/config/.gitignore
@@ -1,0 +1,1 @@
+config.priv.exs
--- a/fission_lib/config/config.secret.exs
+++ b/fission_lib/config/config.secret.exs
@@ -1,0 +1,3 @@
+import Config
+
+config :fission_lib, atomvm_path: "/Users/matheksm/ew/FissionVM/build/src/AtomVM"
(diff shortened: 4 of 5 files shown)


fission_lib/lib/fission_lib.ex
```diff
@@ -15,24 +15,25 @@ defmodule FissionLib do
...
       )
       |> Map.new()
 
     :packbeam_api.create(
       ~c"#{options.output_path}",
-      Enum.map(
-        [options.fission_lib_path] ++ Path.wildcard("#{@build_path}/**/*.{beam,app}"),
-        &String.to_charlist/1
-      ),
+      Enum.map([options.fission_lib_path] ++ options.artifacts, &String.to_charlist/1),
```
jgonet: Nit: `[options.fission_lib_path | options.artifacts]`

fission_lib/patches/estdlib/erl_anno.erl
```diff
@@ -0,0 +1,9 @@
+%% Patch reason: original set_file/2 crashes sometimes ¯\_(ツ)_/¯
+%% for example in "spawn" test in eval_test.exs
+%% See https://linear.app/swmansion/project/erl-annoset-file2-sometimes-crashes-in-erl-eval-76e146323242/overview
+
+-module(erl_anno).
+
+-export([set_file/2]).
+
+set_file(_File, Anno) -> Anno.
```
jgonet, line 9: We may be getting to last branch – filename isn't list or binary thus not settable.

```erlang
set_file(File, Anno) ->
    set(file, File, Anno).

set(Item, Value, Anno) ->
    case {is_settable(Item, Value), Anno} of
        {true, Line} when ?ALINE(Line) ->
            set_anno(Item, Value, [{location, Line}]);
        {true, {L, C} = Location} when ?ALINE(L), ?ACOLUMN(C) ->
            set_anno(Item, Value, [{location, Location}]);
        {true, A} when is_list(A), A =/= [] ->
            set_anno(Item, Value, Anno);
        _ ->
            erlang:error(badarg, [Item, Value, Anno])
    end.
```

fission_lib/config/config.exs
```diff
@@ -1,3 +1,9 @@
 import Config
 
 config :fission_lib, add_tracing: true
```
TheSobkiewicz, line 3: Add atomvm_path with default value to make it easier to track.

mat-hek: I don't think there's a default value for the atomvm path... I added a comment a few lines below

fission_lib/patches/estdlib/erl_anno.erl
mat-hek: Looked at this for a while, not sure why it happens. I'll merge this and we can tackle it in a separate PR

PR #9, opened 2024-11-28
Add simple Elixir REPL

TheSobkiewicz:
Added simple elixir repl. Simple evaluation works.
Patched:
- elixir_sup.erl
- elixir.erl
- elixir.hrl(needed by elixir.erl)

### DIFF
--- a/iex_wasm/lib/elixir_wasm.ex
+++ b/iex_wasm/lib/elixir_wasm.ex
@@ -1,12 +1,13 @@
 defmodule ElixirWasm do
   def start() do
-    :elixir.start([], [])
     spawn(&loop/0)
   end
 
   defp loop() do
     receive do
       {:emscripten, {:call, promise, data}} ->
+        if :elixir_config |> Process.whereis() |> is_nil(), do: :elixir.start([], [])
+
         result =
           case Code.eval_string(data, [], __ENV__) do
             {result, _binding} -> result


fission_lib/patches/estdlib/elixir_sup.erl
jgonet, line 1: Can you add comment what was patched here?

iex_wasm/lib/elixir_wasm.ex
```diff
@@ -0,0 +1,23 @@
+defmodule ElixirWasm do
+  def start() do
+    :elixir.start([], [])
+    :elixir_config.start_link()
+    spawn(&loop/0)
+  end
```
jgonet: Can we keep it in the same file and use common logic for dispatching?

iex_wasm/lib/elixir_wasm.ex
```diff
@@ -0,0 +1,23 @@
+defmodule ElixirWasm do
+  def start() do
+    spawn(&loop/0)
+  end
+
+  defp loop() do
+    receive do
+      {:emscripten, {:call, promise, data}} ->
+        if :elixir_config |> Process.whereis() |> is_nil(), do: :elixir.start([], [])
```
jgonet, line 9: Maybe just `unless Process.whereis(:elixir_config), do: :elixir.start([], [])`

PR #11, opened 2024-11-29
FissionLib: Patch private functions

mat-hek:
- Override private functions by adding `-compile({flb_patch_private, fun_name/arity}).` or `@compile {:flb_patch_private, fun_name: arity}` and defining the function in the patch
  - In Elixir, the overriding functions must be public, otherwise they're pruned away by the compiler. The same can be done in Erlang to avoid an 'unused function' warning, but it should work anyway. The overriding functions are private in the final module regardless of the patch.
- Access all functions (including private ones only present in the original module) via `flb_module` mock module (call `flb_module:foo(args)` instead of `foo(args)`).

Briefly tested so far ;)

### DIFF
--- a/fission_lib/lib/fission_lib/core_erlang_utils.ex
+++ b/fission_lib/lib/fission_lib/core_erlang_utils.ex
@@ -58,7 +58,8 @@ defmodule FissionLib.CoreErlangUtils do
     orig_body =
       orig_body
       |> remove_funs(
-        MapSet.intersection(orig_exports, patch_exports)
+        orig_exports
+        |> MapSet.intersection(patch_exports)
         |> MapSet.union(patch_private_overrides)
       )
       |> rename_funs_and_local_calls(%{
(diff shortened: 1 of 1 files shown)


fission_lib/lib/fission_lib/core_erlang_utils.ex
```diff
@@ -31,26 +31,56 @@ defmodule FissionLib.CoreErlangUtils do
...
+      |> MapSet.difference(patch_private_overrides)
+
+    exports = MapSet.union(orig_exports, patch_exports)
 
     orig_body =
-      rename_funs(orig_body, %{
+      orig_body
+      |> remove_funs(
+        MapSet.intersection(orig_exports, patch_exports)
+        |> MapSet.union(patch_private_overrides)
```
jgonet: Nit: can we pipe `orig_exports` too?

fission_lib/lib/fission_lib/core_erlang_utils.ex
```diff
@@ -93,33 +123,62 @@ defmodule FissionLib.CoreErlangUtils do
...
+        {fun, arity} in funs
+
+      _ast ->
+        false
+    end)
+  end
+
+  defp inject_private_calls(
+         {:c_call, call_meta, {:c_literal, mod_meta, :flb_module}, {:c_literal, fun_meta, fun},
+          args},
```
jgonet: Nit: can we match inside the function to not split it into multiple lines?

PR #12, opened 2024-12-01
Add module compilation

jgonet:
- Added patches to existing `erlang:md5`, `erts_internal:map_next/3`, `lists: duplicate/2`
- Added patches for `compile.erl` and `beam_ssa_pre_codegen.erl` (compression, [exceeding literal list allocation](https://github.com/atomvm/AtomVM/issues/1379))
- Added support for module compilation in frontend and main app module
- Recompiled WASM to jgonet/swm/compile@bb7d704c (will add PR for changes shortly).
- Added .gitignore for DS_Store (which is annoying)

image

Reviewers: @TheSobkiewicz, @mat-hek

### DIFF
--- a/fission_lib/patches/estdlib/erlang.erl
+++ b/fission_lib/patches/estdlib/erlang.erl
@@ -310,9 +310,9 @@ system_flag(_Key, _Value) ->
 %% @end
 %%-----------------------------------------------------------------------------
 -spec md5(Data :: binary()) -> binary().
-%% Patch reason: md5 uses native mbedtls implementation that isn't easily linked
-%so we use vendored implementation
-%% Patch reason: md5 accepts iolist as well
+%% Patch reason: AVM patch; md5 uses native mbedtls implementation that isn't easily linked
+%% so we use vendored implementation
+%% Patch reason: AVM patch; md5 accepts iolist as well
 md5(Data) when is_binary(Data) orelse is_list(Data) ->
     erlang:nif_error(undefined).
 
--- a/fission_lib/patches/estdlib/lists.erl
+++ b/fission_lib/patches/estdlib/lists.erl
@@ -736,7 +736,7 @@ unique([X, Y | Tail], Fun) ->
 %% @end
 %%-----------------------------------------------------------------------------
 -spec duplicate(integer(), Elem) -> [Elem].
-%% Patch reason: Count can be 0 too
+%% Patch reason: AVM patch; Count can be 0 too
 duplicate(Count, Elem) when is_integer(Count) andalso Count >= 0 ->
     duplicate(Count, Elem, []).
 
--- a/iex_wasm/lib/iex_wasm.ex
+++ b/iex_wasm/lib/iex_wasm.ex
@@ -45,7 +45,7 @@ defmodule IExWASM do
         |> Enum.map(parse_form)
         |> :compile.noenv_forms(compile_opts)
 
-      :code.load_binary(module, "nofile", module_bin)
+      :code.load_binary(module, ~c"nofile", module_bin)
     catch
       error -> {:error, error}
     end


fission_lib/patches/estdlib/beam_asm.hrl
mat-hek, line 1: would be great not to vendor these hrls, but no idea how

fission_lib/patches/estdlib/compile.erl
mat-hek, line 1: Have you tried patching only the changed functions?

fission_lib/patches/estdlib/erlang.erl
```diff
@@ -310,8 +310,11 @@ system_flag(_Key, _Value) ->
 %% @end
 %%-----------------------------------------------------------------------------
 -spec md5(Data :: binary()) -> binary().
-md5(Data) when is_binary(Data) ->
-    crypto:hash(md5, Data).
+%% Patch reason: md5 uses native mbedtls implementation that isn't easily linked
+%so we use vendored implementation
+%% Patch reason: md5 accepts iolist as well
```
mat-hek:
```suggestion
```

Patch reasons are meant to describe differences from the stdlib, not the AtomVM version

fission_lib/patches/estdlib/lists.erl
```diff
@@ -736,7 +736,8 @@ unique([X, Y | Tail], Fun) ->
 %% @end
 %%-----------------------------------------------------------------------------
 -spec duplicate(integer(), Elem) -> [Elem].
-duplicate(Count, Elem) when is_integer(Count) andalso Count > 0 ->
+%% Patch reason: Count can be 0 too
```
mat-hek:
```suggestion
```

iex_wasm/lib/iex_wasm.ex
```diff
@@ -1,33 +1,84 @@
...
+      {:dot, _} = f, {current, forms} ->
+        new_form = Enum.reverse([f | current])
+        {[], [new_form | forms]}
+
+      f, {current, forms} ->
+        {[f | current], forms}
+    end
+
+    {[], split} = Enum.reduce(forms, {[], []}, split_on_dots)
+    Enum.reverse(split)
```
mat-hek:
```suggestion
    split_on_dots = fn
      {:dot, _} = f, current -> {[Enum.reverse([f | current])], []}
      f, current -> {[], [f | current]}
    end

    {split, []} = Enum.flat_map_reduce(forms, [], split_on_dots)
    split
```

iex_wasm/lib/iex_wasm.ex
```diff
@@ -1,33 +1,84 @@
...
+    try do
+      {:ok, tokens, _} = :erl_scan.string(string)
+
+      {:ok, module, module_bin} =
+        tokens
+        |> split_forms()
+        |> Enum.map(parse_form)
+        |> :compile.noenv_forms(compile_opts)
+
+      :code.load_binary(module, "nofile", module_bin)
```
mat-hek: Technically it should be

```suggestion
      :code.load_binary(module, ~c"nofile", module_bin)
```

fission_lib/patches/estdlib/beam_types.hrl
mat-hek, line 1: what has changed in this file?

jgonet: Nothing IIRC, we need hrl for successful compilation.

fission_lib/patches/estdlib/compile.erl
jgonet: Yeah, that may be a good idea. Not sure if worth though – compile functions heavily depend on each other and I wanted to improve this after we have private functions patching.

fission_lib/patches/estdlib/erlang.erl
jgonet: Eh, I think we should have some markers for any patch so we don't lose track of _why_ we need the given patch. Do you want to use other naming?

iex_wasm/lib/iex_wasm.ex
jgonet: I don't understand the use of `Enum.flat_map_reduce` here.

mat-hek: Here instead of manually storing the result in the 'forms' part of the accumulator, it's done in the `flat_map` part. Now I think it would be best done with `Enum.chunk_while`. But you can leave it as is if you like.

fission_lib/patches/estdlib/beam_types.hrl
mat-hek: ok I missed the fact it's hrl

iex_wasm/static/script.js
```diff
@@ -39,17 +66,16 @@ const Elements = {
...
   };
   Elements.clearButton.onclick = () => {
     Elements.logsDisplay.innerHTML = "";
   };
-  Elements.evalButton.onclick = () => evalCode();
+  Elements.evalButton.onclick = () => evalErlang();
+  Elements.moduleButton.onclick = () => compileErlangModule();
   Elements.codeInput.addEventListener("keydown", (event) => {
     const cmdEnter = event.key === "Enter" && (event.metaKey || event.ctrlKey);
     if (cmdEnter) {
```
TheSobkiewicz, line 81: evalCode no longer exists.

iex_wasm/lib/iex_wasm.ex
jgonet: I missed `Enum.chunk_while` when looking for good function to do that, it's much better.

PR #18, opened 2024-12-05
Cache patches

mat-hek:


fission_lib/lib/fission_lib/build.ex
```diff
@@ -1,116 +1,180 @@
...
 
+  @config FissionLib.Config.get([:add_tracing, :ex_stdlib_beam_paths, :erl_stdlib_beam_paths])
+
+  @app_path Mix.Project.app_path()
+  @build_path "#{@app_path}/flb_patches_ebin"
+  @cache_path "#{@app_path}/flb_patch_cache"
+
+  # When this module is recompiled, we need to
+  # patch from scratch
+  File.rm(@cache_path)
```
jgonet, line 14: This is clever.

fission_lib/lib/fission_lib/build.ex
```diff
@@ -1,116 +1,180 @@
...
 
   defp process_async(enum, fun, opts \\ []) do
     enum
     |> Task.async_stream(fun, [timeout: 30_000, ordered: false] ++ opts)
     |> Stream.run()
   end
+
+  defp with_tmp_dir(path, fun) do
+    tmp_dir = "#{path}/tmp_#{:erlang.unique_integer([:positive])}"
+    File.rm_rf!(tmp_dir)
```
jgonet, line 175: Shouldn't we remove wildcard of `tmp_`?

mat-hek: Then we could remove other folders that are created in parallel 🤔 But indeed some tmp_dirs that survived in previous runs may make it into the bundle. I think we should remove the cache every time on startup, in case something breaks. It will fix other potential problems too.

jgonet: My point is that we generate unique path that doesn't yet exist and then we try to delete it, as far as I understand the code.

fission_lib/lib/fission_lib/build.ex
mat-hek: 🧠

fission_lib/lib/fission_lib/build.ex
mat-hek: If executing `fun` breaks, or if you do `mix compile` and hit `ctrl+c`, the `tmp_dir` may not be removed. Then when you do `mix compile` again, the 'unique path' may already exist.

fission_lib/lib/fission_lib.ex
```diff
@@ -3,9 +3,12 @@ defmodule FissionLib do
   Library providing Erlang and Elixir stdlibs, allowing
   to compile projects to `.avm` and run them with AtomVM.
   """
-  alias __MODULE__.Config
+  require __MODULE__.Config
 
   @build_path Mix.Project.build_path()
+  @app_path Mix.Project.app_path()
+
+  @config __MODULE__.Config.get([:start_module, :out_path])
```
jgonet: Can we use `FissionLib.Config` instead of `__MODULE__`? `FissionLib.Build` does it this way and IMO it's more readable.

fission_lib/lib/fission_lib/build.ex
```diff
@@ -1,116 +1,181 @@
...
+           true <- Enum.empty?(missing_srcs) do
+        cache
+      else
+        _cant_use_cache ->
+          File.rm_rf!(@build_path)
+          File.mkdir_p!(@build_path)
+          %{}
+      end
+
+    File.rm_rf!(@cache_path)
```
jgonet: We should be able to go with `File.rm`, right? It's a single file.

PR #21, opened 2024-12-11
Update WASM to 7cce6378

jgonet:
@TheSobkiewicz, @mat-hek

### DIFF
--- a/fission_lib/patches/estdlib/lists.erl
+++ b/fission_lib/patches/estdlib/lists.erl
@@ -58,9 +58,11 @@
     seq/2, seq/3,
     sort/1, sort/2,
     split/2,
+    partition/2,
     usort/1, usort/2,
     duplicate/2,
-    sublist/2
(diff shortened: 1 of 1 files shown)


fission_lib/patches/estdlib/lists.erl
```diff
@@ -674,6 +676,24 @@ split(N, [H | T], R) ->
 split(_, [], _) ->
     badarg.
 
+partition(Pred, L) when is_function(Pred, 1) ->
+    partition(Pred, L, [], []).
+
+partition(_Pred, [], A, B) ->
+    {A, B};
```
TheSobkiewicz, line 683: Original implementation uses reverse here, might be worth adding .

mat-hek: +1, even if docs don't say that, I wouldn't be surprised if something relied on that, and we'll be the ones debugging

PR #22, opened 2024-12-18
Cleaned compile and elixir files

TheSobkiewicz:
Cleaned compile.erl and elixir.erl files.


fission_lib/patches/estdlib/compile.erl
```diff
@@ -223,1633 +56,12 @@ env_default_opts() ->
...
-                            {error, [{none, ?MODULE, no_crypto_key}]};
-                        Key ->
-                            encrypt_debug_info(DebugInfo, {Mode, Key}, Opts2)
-                    end
-            end;
-        false ->
-            {ok, DebugInfo, Opts2}
-    end.
-
 debug_info_chunk(#compile{
```
mat-hek, line 59: missing patch reason

TheSobkiewicz: image

PR #23, opened 2024-12-20
fission lib: add stacktrace test

mat-hek:


fission_lib/test/support/run_in_atomvm.ex
```diff
@@ -60,6 +60,7 @@ defmodule RunInAtomVM do
         end
 
         defp run(opts) do
+          _no_warn_unused = opts
```
jgonet, line 63:
```suggestion
        defp run(_opts) do
```

mat-hek: This is on purpose, otherwise if there's a warning when we don't use opts in the evaluated code

PR #24, opened 2024-12-23
Remove AtomVM.{js, wasm} from iex_wasm

mat-hek:


fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+      Process.sleep(:infinity)
+    end
+
+    :ok
+  end
+
+  defp cp_artifact(build_dir, subpath, options) do
+    File.cp!(
+      Path.join(build_dir, subpath),
+      Path.join(options.out_dir, "#{options.out_name}#{Path.extname(subpath)}")
```
jgonet: Nit: `options.out_name <> Path.extname(subpath)`.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+    System.shell("#{cmd} 1>&2")
+    |> handle_shell_status()
+  end
+
+  defp handle_shell_status({_out, 0}), do: :ok
+
+  defp handle_shell_status({_out, status}) do
+    if status != 0 do
+      System.stop(status)
+      Process.sleep(:infinity)
```
jgonet: Why we need this sleep?

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+
+  defp shell(cmd) do
+    System.shell("#{cmd} 1>&2")
+    |> handle_shell_status()
+  end
+
+  defp handle_shell_status({_out, 0}), do: :ok
+
+  defp handle_shell_status({_out, status}) do
+    if status != 0 do
```
jgonet: Isn't status always non-zero here?

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+  Firstly, configure your project with
+
+  ```
+  config :fission_lib, avm_source: <source>
+  ```
+
+  where the <source> can be:
+  - `{:path, string}` - local path to AtomVM source
+  - `{:git, address}` - GIT address to clone the AtomVM from
+  - `{:git, address, branch: branch}` - like above, allows specifying
```
jgonet: Can we use `ref` instead of `branch` name? It typically covers all three.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+      Please configure AtomVM source with
+
+      config :fission_lib, avm_source: "path/to/atomvm"
+      """
+    end
+
+    defaults = %{target: :unix, out_dir: ".", out_name: "AtomVM"}
+
+    {options, _rest} =
+      OptionParser.parse!(args, strict: defaults |> Map.keys() |> Keyword.from_keys(:string))
```
jgonet: Can we extract parser config to variable? Also limits Elixir to min 1.14 (`Keyword.from_keys`).

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+  def run(args) do
+    unless @config.avm_source do
+      raise """
+      Please configure AtomVM source with
+
+      config :fission_lib, avm_source: "path/to/atomvm"
+      """
+    end
+
+    defaults = %{target: :unix, out_dir: ".", out_name: "AtomVM"}
```
jgonet: Can be module attribute.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+
+    pwd = File.cwd!()
+
+    case String.to_existing_atom(options.target) do
+      :unix ->
+        build_dir = Path.join(avm_source, "build")
+        File.mkdir_p!(build_dir)
+        File.cd!(build_dir)
+        shell("cmake ..")
+        shell("make -j #{System.schedulers()}")
```
jgonet: Don't need to pass number to `-j`, it uses [all available threads](https://www.gnu.org/software/make/manual/html_node/Options-Summary.html).

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+        File.cd!(pwd)
+        cp_artifact(build_dir, "src/AtomVM", options)
+
+      :wasm ->
+        build_dir = Path.join(avm_source, "src/platforms/emscripten/build")
+        File.mkdir_p!(build_dir)
+        File.cd!(build_dir)
+        shell("emcmake cmake .. -DAVM_EMSCRIPTEN_ENV=web")
```
jgonet: ⚠️ Doesn't it have dependency on main cmake being executed? It defines some variables used in C source.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+    IO.puts("Cloning AtomVM from #{addr}")
+
+    System.cmd("git", ["clone", addr] ++ branch ++ ["--", output])
+    |> handle_shell_status()
+
+    output
+  end
+
+  defp shell(cmd) do
+    System.shell("#{cmd} 1>&2")
```
jgonet: Why we want to redirect stderr as well?
I wonder if passing everything through `System.cmd` wouldn't be simpler – we could inline shell status handling and maybe even directory switching.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+        File.cd!(pwd)
+        cp_artifact(build_dir, "src/AtomVM.js", options)
+        cp_artifact(build_dir, "src/AtomVM.wasm", options)
+    end
+  end
+
+  defp fetch_repo(addr, opts \\ []) do
+    branch = if opts[:branch], do: ["-b", opts[:branch]], else: []
+    output = Path.join(@build_dir, "atomvm_src")
+    File.rm_rf!(output)
```
jgonet, line 75: ❗ I won't do that automatically. If someone makes an error then you possibly silently clobber entire fs subtree.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
```diff
@@ -0,0 +1,112 @@
...
+    IO.puts("Cloning AtomVM from #{addr}")
```
jgonet: ❗ We should always print to stderr. Here and in the patcher as well.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: Because `System.stop` is async

> Note this function is asynchronous and the current process will continue
executing after this function is invoked. In case you want to block the current
process until the system effectively shuts down, you can invoke
Process.sleep(:infinity).

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: > Also limits Elixir to min 1.14

Now we only support 1.17 :D I'm not sure if it's worth limiting ourselves until we at least decide which versions we want

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: I have no idea what happens in this cmake spaghetti, and I'd rather avoid getting into that :D Just following the instructions [here](https://www.atomvm.net/doc/v0.6.0-alpha.0/build-instructions.html#building-for-the-web). But I can run cmake in the root directory just in case :P

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: Can you elaborate? What kind of error?

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: Even if Mix prints to stdout?

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: The point here is to avoid the command output being captured by `System.shell`. Just figured out we can pass `use_stdio: false` too.

> we could inline shell status handling

How would we do that?

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
jgonet: I'd rather check if it builds on a cleanly cloned repo.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
jgonet: It doesn't (see below). It's a common practice to output user facing logs to stderr.
```
➜  iex_wasm git:(main) ✗ mix compile
Bundling fission_lib.avm
    warning: function loop/0 is unused
    │
 24 │   defp loop() do
    │        ~
    │
    └─ lib/elixir_wasm.ex:24:8: ElixirWasm (module)

    warning: :emscripten.promise_resolve/2 is undefined (module :emscripten is not available or is yet to be defined)
    │
 36 │         :emscripten.promise_resolve(promise, result_str)
    │                     ~
    │
    └─ lib/elixir_wasm.ex:36:21: ElixirWasm.loop/0

    warning: Console.print/1 is undefined (module Console is not available or is yet to be defined)
    │
  3 │     Console.print("Starting application...\n")
    │             ~
    │
    └─ lib/iex_wasm.ex:3:13: IExWASM.start/0

    warning: :emscripten.promise_resolve/2 is undefined (module :emscripten is not available or is yet to be defined)
    │
 70 │     :emscripten.promise_resolve(promise, value)
    │                 ~
    │
    └─ lib/iex_wasm.ex:70:17: IExWASM.resolve/2

➜  iex_wasm git:(main) ✗ mix compile 2>/dev/null
Bundling fission_lib.avm
```

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
jgonet: Actually, nevermind – we only remove `$BUILD_PATH/atomvm_src`. I was thinking about user configuring directories in the wrong place (e.g. `~/`) and we'll be removing user data without confirmation.

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
jgonet: Something like that:
```elixir
  defp cmd(cmd, opts \\ []) do
    [command | args] = cmd
    opts = Keyword.put(opts, :use_stdio, false)

    case System.cmd(command, args, opts) do
      {_output, 0} ->
        nil

      {_output, status} ->
        System.stop(status)
        Process.sleep(:infinity)
    end

    :ok
  end
```

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: Not always, but ok

```
mix deps.get 2>/dev/null
* Getting atomvm_packbeam (https://github.com/atomvm/atomvm_packbeam.git)
remote: Enumerating objects: 358, done.
remote: Counting objects: 100% (153/153), done.
remote: Compressing objects: 100% (102/102), done.
remote: Total 358 (delta 68), reused 119 (delta 51), pack-reused 205 (from 1)
```

fission_lib/lib/mix/tasks/fission_lib.build_avm.ex
mat-hek: How would that help?

jgonet: My ideal usecase is to download this repo, clone (or do it automatically if more convenient) AVM repo and build wasm without touching AVM repo. I'd not run main `cmake` / `make` if not necessary for WASM builds but I didn't check if it build WASM without that so that's the the thing I'm concerned about.

mat-hek: So WASM seems to build & work without running the main cmake, but I'm unsure if it changes anything. I also don't see how running the main cmake could break something though.

jgonet: I was wondering if it's necessary to run. We're good to go.

PR #32, opened 2025-01-27
Add release instructions

jgonet:


iex_wasm/README.md
```diff
@@ -25,8 +25,17 @@ elixir server.exs
...
 ```
+
 for the WASM AtomVM to work.
+
+## Deploy
+
+- Build `.avm` file and AtomVM WASM binary in release mode (`MIX_ENV=prod mix compile`).
+- Connect to server via ssh.
+- Copy files into the server (`scp -A -r <SOURCE_DIR> root@<IP>:/var/www/wasm-$(date "+%Y-%m-%d")`).
+- Update symlink to point to new directory (in `/var/www/`: `ln -sf ./wasm-DATE html`)
```
mat-hek, line 41: Awesome, can it become `mix deploy`? :P

jgonet: I don't see why not, I'll experiment with it.

PR #33, opened 2025-01-28
Refactor Elixir handling in WASM

jgonet:
## Changes

- Simplify JS
- Make Elixir the default
- Remove separate entrypoint for Elixir, rename the entrypoint to `App`.

cc: @mat-hek, @TheSobkiewicz 

<details><summary><h2>Screenshots<h2></summary>
<p>

## New Elixir

image

## New Erlang

image

## Narrow Erlang page

image

</p>
</details>

### DIFF
--- a/iex_wasm/lib/app.ex
+++ b/iex_wasm/lib/app.ex
@@ -46,8 +46,8 @@ defmodule App do
       form
     end
 
-    with code = :erlang.binary_to_list(code),
-         {:ok, tokens, _end_location} <- :erl_scan.string(code),
+    code = :erlang.binary_to_list(code)
+    with {:ok, tokens, _end_location} <- :erl_scan.string(code),
          {:ok, module, module_bin} <-
            tokens
            |> split_forms()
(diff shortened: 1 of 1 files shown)


iex_wasm/lib/app.ex
```diff
@@ -38,7 +46,8 @@ defmodule IExWASM do
       form
     end
 
-    with {:ok, tokens, _end_location} <- :erl_scan.string(string),
+    with code = :erlang.binary_to_list(code),
```
mat-hek: can be moved out of `with`

iex_wasm/lib/app.ex
```diff
@@ -51,8 +60,9 @@ defmodule IExWASM do
     error -> {:error, error, __STACKTRACE__}
   end
 
-  defp eval(string, :eval) do
-    with {:ok, tokens, _end_location} <- :erl_scan.string(string),
+  defp eval(code, :erlang) do
+    with code = :erlang.binary_to_list(code),
```
mat-hek: can be moved out of `with`

iex_wasm/static/erlang.html
mat-hek, line 1: Can we remove this file and just have a radio selector with the language?

jgonet: Ehh, we can but we'd need to do much more work – replacing content dynamically in JS. I don't expect html to change much so I'd just leave it as is.

PR #34, opened 2025-01-31
Precompile tests before running

jgonet:
- Allows to precompile eval tests since it uses the same compiled .avm every time
- Add `eval/2` helper and helper assertions
- Run all tests concurrently using Boombox macro

To allow running tests concurrently, we use env variable on each test run – this is blocked by https://github.com/software-mansion-labs/FissionVM/pull/90.

### DIFF
--- a/fission_lib/test/support/atomvm.ex
+++ b/fission_lib/test/support/atomvm.ex
@@ -223,7 +223,7 @@ defmodule FissionLib.Support.AtomVM do
         @compile autoload: false, no_warn_undefined: :atomvm
 
         def start() do
-          run_dir = :atomvm.posix_getenv(~c"AVM_RUN_DIR")
+          run_dir = :os.getenv(~c"AVM_RUN_DIR")
           args = read_args(run_dir)
           result = run(args)
           write_result(result, run_dir)


fission_lib/test/support/atomvm.ex
```diff
@@ -0,0 +1,262 @@
...
+  when this module runs `start/0`. We then write code output to `result.bin`.
+
+  Compiled files are cached with phash2 key of AST contents under app's priv directory.
+  All eval tests run the same underlying code so `eval` expects
+  module to be compiled and cached prior to call.
+  """
+  import ExUnit.Assertions
+
+  @atomvm_path Application.compile_env!(:fission_lib, :atomvm_path)
+  @compile_dir :code.priv_dir(:fission_lib)
```
jgonet: They're semi-persistent so we can store them at `priv/`. Logs are stored there as well for now, I may reconsider storing logs and input/output bin files in `/tmp` elsewhere later.

fission_lib/lib/fission_lib.ex
```diff
@@ -34,7 +34,10 @@ defmodule FissionLib do
       )
       |> Map.new()
 
-    fission_lib_artifacts = Path.wildcard("#{@app_path}/**/*")
+    # don't include priv directory
+    fission_lib_priv = Path.wildcard("#{@app_path}/**/priv/**/*")
+    fission_lib_artifacts = Path.wildcard("#{@app_path}/**/*") -- fission_lib_priv
+
     artifacts = [options.fission_lib_path | options.artifacts -- fission_lib_artifacts]
```
mat-hek: This way fission_lib's priv directory is included in the bundle, is that what you wanted? Why so?

fission_lib/test/support/atomvm.ex
```diff
@@ -1,62 +1,234 @@
...
+    run_id = :erlang.unique_integer([:positive])
+
+    result_path = Path.join(dir, "result-#{run_id}.bin")
+    bindings_path = Path.join(dir, "opts-#{run_id}.bin")
+
+    logs_dir = Path.join(dir, "logs")
+    File.mkdir_p!(logs_dir)
+
+    unix_now = DateTime.now!("Etc/UTC") |> DateTime.to_iso8601()
+    log_path = Path.join(logs_dir, "#{unix_now}.txt")
```
mat-hek: This way we don't know where are the logs from a particular test, and the file path changes each time we run the test. Currently, I can keep the log file open and its path indicates which test generated it, which is convenient. I think we need to change it before merging.

fission_lib/test/support/atomvm.ex
```diff
@@ -1,62 +1,234 @@
...
+    {_output, 0} = System.shell("elixirc #{source_path} -o #{build_dir}")
+
+    beams = build_dir |> Path.join("*.beam") |> Path.wildcard()
+    paths = Enum.map(beams, copy_beam_to_dir)
+    File.rm_rf!(build_dir)
+
+    paths
+  end
+
+  def eval_fragment(:elixir) do
```
mat-hek: The name suggests that this actually evaluates code, while it only creates AST

fission_lib/test/support/atomvm.ex
```diff
@@ -1,62 +1,234 @@
 defmodule FissionLib.AtomVM do
```
mat-hek:
```suggestion
defmodule FissionLib.Support.AtomVM do
```

To distinguish from non-test modules

fission_lib/test/support/test_async.ex
```diff
@@ -0,0 +1,57 @@
+defmodule FissionLib.AsyncTest do
```
mat-hek:
```suggestion
defmodule FissionLib.Support.AsyncTest do
```

fission_lib/test/support/atomvm.ex
jgonet: It's ambiguous, will change. The intent was "something_fragment" as in ast fragment.

fission_lib/test/fission_lib/stacktrace_test.exs
```diff
@@ -1,17 +1,24 @@
...
       quote do
         try do
           :orddict.take(:not_an_orddict, 1)
         rescue
           e -> {e, __STACKTRACE__}
         end
       end
-      |> RunInAtomVM.expr(tmp_dir)
+      |> AtomVM.compile_quoted([])
+      |> AtomVM.run_with_bindings(run_dir, [])
```
mat-hek: - I think there was a function that did compile and run_with_bindings, which didn't require specifying the bindings twice. If we don't want that, I think it's better to have a single binding like `args` and pass a map in there.
- An empty list could be a default

jgonet: > I think there was a function that did compile and run_with_bindings, which didn't require specifying the bindings twice.

It did compile and run in one step. Most of our tests are evaluation tests so I assumed we can have two low level functions that `eval` builds upon instead of rarely used `compile_and_run_quoted`.

> If we don't want that, I think it's better to have a single binding like args and pass a map in there.

Kwlist vs map doesn't matter here IMO but I can change it. BTW, this list contains bindings that can be set to change values after code is compiled. So, the list is `args`. If we don't pass anything then we're running code that's static after compilation.

> An empty list could be a default
I don't like this, since the function name is `AtomVM.run_with_bindings` so we don't want to hide the mentioned bindings.

mat-hek: >  I assumed we can have two low level functions that eval builds upon

`eval` doesn't build upon these functions, it just calls `run_with_bindings`

> instead of rarely used compile_and_run_quoted

It's rarely used because now we focus on compilation, but generally it's the most basic case - compile code and run in AVM

> Kwlist vs map doesn't matter here IMO 

The point is that instead of

```elixir
      quote do
        var!(n) + 3
      end
      |> AtomVM.compile_quoted([:n])
      |> AtomVM.run_with_bindings(run_dir, n: 2)
```

you could do

```elixir
      quote do
        args.n + 3
      end
      |> AtomVM.compile_quoted()
      |> AtomVM.run_with_bindings(run_dir, n: 2)
```

> we don't want to hide the mentioned bindings.

Then maybe it should be `run`?

jgonet: > The point is that instead (...)

Ah, I see. It makes sense, along with `run` suggestion, will change.

PR #36, opened 2025-02-03
Fuzzer improvements

jgonet:
cc: @TheSobkiewicz, @mat-hek

### DIFF
--- a/fuzzer/run-atomvm-test.sh
+++ b/fuzzer/run-atomvm-test.sh
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 main() {
-  atomvm_timeout_in_s="${AVM_FUZZ_TIMEOUT_S}"
+  atomvm_timeout_in_s="${AVM_FUZZ_TIMEOUT_S:-5}"
   atomvm_path="${AVM_ATOM_VM_BIN}"
   packbeam_path="${AVM_PACKBEAM_BIN}"
   avm_lib_path="${AVM_FISSION_LIB_AVM}"
--- a/fuzzer/README.md
+++ b/fuzzer/README.md
@@ -13,7 +13,7 @@
 
 - `cargo build --release` in `erlfuzz/`
 - create `.env` file with three environment variables containing absolute paths to needed binaries
-  - `AVM_FUZZ_TIMEOUT_S` – sets maximum time runner will wait until sample execution completes
+  - `AVM_FUZZ_TIMEOUT_S` (optional, default 5s) – sets maximum time runner will wait until sample execution completes
   - `AVM_ERLFUZZ_BIN` – path to `erlfuzz` binary
   - `AVM_ATOM_VM_BIN` – path to `AtomVM` binary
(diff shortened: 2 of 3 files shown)


fuzzer/error-segregation.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if [[ -z "$1" ]]; then
+  echo "Usage: $0 <path_to_directory>"
+  exit 1
+fi
+
+directory="$1"
+
+mkdir -p "${directory}/nbits"
+mkdir -p "${directory}/asan"
```
TheSobkiewicz, line 21: I recommend to add another folder for AddressSanitizer: heap-use-after-free

fuzzer/error-segregation.sh
```diff
@@ -0,0 +1,63 @@
...
+mkdir -p "${directory}/operand"
+mkdir -p "${directory}/compilation_error"
```
TheSobkiewicz, line 23: Add folder for "Unknown external term type"

fuzzer/run-atomvm-test.sh
```diff
@@ -1,32 +1,33 @@
 #!/bin/bash
-set -uo pipefail
+set -euo pipefail
+
 main() {
-  atomvm_timeout_in_s=30
-  atomvm_path=""
-  packbeam_path=""
-  avm_lib_path="../fission_lib/_build/dev/fission_lib.avm" 
+  atomvm_timeout_in_s="${AVM_FUZZ_TIMEOUT_S}"
```
TheSobkiewicz: Not sure if this really need to be env variable, maybe add default value

jgonet: It's easier to configure that way. I use 5s timeout instead of default 30s.

fuzzer/error-segregation.sh
jgonet: I got only one report for that in 1k runs. Other added are more common. Same applies to unknown external term type.

TheSobkiewicz: Okay, we will have to change it after merging https://github.com/software-mansion-labs/erlfuzz/pull/1 anyway, according to the fact that it completely changes the error distribution

fuzzer/run-atomvm-test.sh
TheSobkiewicz: Okay, so just leave 30 sec as default option

jgonet: cc @mat-hek for tie-break since I'm with opinion that timeout should be explicit instead of optional for visibility.

mat-hek: IMO let's make the timeout configurable and set the default to 5s :D

mat-hek: Plus let's make sure that we have a proper error if the timeout fires, that mentions how to configure the timeout

jgonet: We can leave a message when we sort samples that were killed during execution – we move them to their own dir anyway.

PR #38, opened 2025-02-06
Delete lists:duplicate/2 patch

FranekKubis:


fission_lib/test/fission_lib/elixir_module_test.exs
```diff
@@ -35,4 +35,10 @@ defmodule FissionLib.ElixirModuleTest do
     File.read!("#{@examples_path}/CapybaraHabitat.ex")
     |> run(tmp_dir)
   end
+
+  test "lists_duplicate" do
+    assert [] = :lists.duplicate(0, "a")
+    assert ["a"] = :lists.duplicate(1, "a")
+    assert ["a", "a", "a", "a", "a", "a", "a"] = :lists.duplicate(7, "a")
+  end
```
jgonet: You need to run it through AtomVM via `run` here, this currently tests OTP stdlib which has correct behavior.

PR #40, opened 2025-02-06
Make apps startup work

mat-hek:
requires https://github.com/software-mansion-labs/FissionVM/pull/100

### DIFF
--- a/fission_lib/test/fission_lib/app_test.exs
+++ b/fission_lib/test/fission_lib/app_test.exs
@@ -1,21 +1,19 @@
 defmodule FissionLib.AppTest do
   use ExUnit.Case, async: true
 
-  alias FissionLib.Support.AtomVM
+  require FissionLib.Support.AtomVM, as: AtomVM
 
   @moduletag :tmp_dir
 
   @tag :app
   test "app", %{tmp_dir: tmp_dir} do
-    info =
-      quote do
-        :application_controller.start(:kernel)
-        :application.ensure_all_started(:elixir)
-        :ok
-      end
-      |> AtomVM.compile_quoted()
-      |> AtomVM.run(tmp_dir)
-
-    assert info.result == :ok
+    quote do
+      :application_controller.start(:kernel)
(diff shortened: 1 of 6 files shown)


fission_lib/patches/eavmlib/init.erl
```diff
@@ -0,0 +1,8 @@
+-module(init).
+
+-export([get_arguments/0, get_argument/1]).
+
+% Patch reason: this is a mock implementation,
+% sufficient for some use cases not to break
```
jgonet, line 6: Please add tickets in linear for every new patch that can be reasonably implemented.

fission_lib/test/fission_lib/app_test.exs
```diff
@@ -0,0 +1,21 @@
...
+    info =
+      quote do
+        :application_controller.start(:kernel)
+        :application.ensure_all_started(:elixir)
+        :ok
+      end
+      |> AtomVM.compile_quoted()
+      |> AtomVM.run(tmp_dir)
+
+    assert info.result == :ok
```
jgonet: Alternative: `|> AtomVM.assert_result(:ok)`

fission_lib/test/test_helper.exs
```diff
@@ -1,5 +1,7 @@
 alias FissionLib.Support.AtomVM
 
+File.rm_rf!("tmp")
```
jgonet, line 4: Do we want to clear cache before every run? It should be pretty much static and matters only when you're changing eval test code.

mat-hek: > matters only when you're changing eval test code

I don't think so. I suppose this should also be recompiled if we change any patch. Anyway, we should probably figure out something better than this

jgonet: Offline: for now we do this, may come back to it in the future.

fission_lib/patches/eavmlib/init.erl
mat-hek: done

PR #42, opened 2025-02-07
Add minimize script, fix error checking in fuzz script

jgonet:
Blocked by https://github.com/software-mansion-labs/erlfuzz/pull/2.

### DIFF
--- a/fuzzer/minimize.sh
+++ b/fuzzer/minimize.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+set -euo pipefail
+
+# Usage: ./minimize.sh <test case name> | <directory ending with slash> | all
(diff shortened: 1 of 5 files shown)


fuzzer/minimize.sh
```diff
@@ -0,0 +1,55 @@
...
+  exit 1
+fi
+
+# automatically export all variables
+set -a
+source .env
+set +a
+
+if [[ "$#" -eq 0 ]]; then
+  echo "Please pass sample name, directory with slash or 'all'"
```
TheSobkiewicz, line 17: It looks okay, I would probably prefer to set all as the default option, so ./minimize.sh without args will reduce every erl file in the folder.  It is just a suggestion, though.

PR #43, opened 2025-02-11
Separate patches and AVM overrides

jgonet:
Mostly for code organization and separating when we needed to patch and when we only used AVM version of stdlib module.

cc: @TheSobkiewicz, @FKubisSWM


mat-hek (changes requested):
I see https://github.com/software-mansion-labs/elixir-wasm/blob/jgonet/categorize-patches/fission_lib/patches/estdlib/lists.erl that doesn't have real patches (only fixes for the AtomVM implementation) and https://github.com/software-mansion-labs/elixir-wasm/blob/jgonet/categorize-patches/fission_lib/patches/estdlib/maps.erl that should also be moved. Please check if other files can be moved as well, especially those starting with the AtomVM copyright notice

PR #47, opened 2025-02-25
Fix greetings.erl example

FranekKubis:


fission_lib/patches/estdlib/erl_parse.erl
```diff
@@ -0,0 +1,30 @@
+%% There is a possibility of ill-implemented single catch command that is why this patch is implemented.
```
jgonet: Add "Patch reason:" please. And we know for sure that catch doesn't work so we can just say it more confidently :P

fission_lib/test/examples/greetings.erl
```diff
@@ -2,61 +2,61 @@
...
-Pid2 ! {self(), request},
-Pid3 ! {self(), request},
-
-% Collect results and display the final message
-receive {Pid1, Language} ->
-    receive {Pid2, Task} ->
-        receive {Pid3, Compliment} ->
-            erlang:display(User ++ ", your mission, if you choose to accept it, is to use "++ Language ++ " to " ++ Task ++ ". Also, " ++ Compliment)
+    User = "Jose",
+
```
jgonet, line 6: If you used some smaller repro, you can add it inline to tests.

FranekKubis: do you mean to add a test case to "erlang_module_test.exs" similar to "Adder"  for example?

jgonet: Yup!

FranekKubis: done - you can check

jgonet: Can we please do it inline instead of creating new file for it?

FranekKubis: done

PR #49, opened 2025-03-03
Check "getting started"/cheatsheet Elixir snippets in WASM

FranekKubis:


fission_lib/test/support/test_async.ex
```diff
@@ -59,4 +61,34 @@ defmodule FissionLib.Support.AsyncTest do
       Module.delete_attribute(__MODULE__, :tag)
     end
   end
+
+  defmacro assert_eval(code_string, expected) do
```
jgonet: Those shouldn't live here, they're specific to getting started snippets.

PR #52, opened 2025-03-06
Partially enable warnings

mat-hek:


fission_lib/patches/estdlib/code.erl
```diff
@@ -98,3 +100,11 @@ load_binary(_Module, _Filename, _Binary) ->
     Module :: atom().
 ensure_loaded(_Module) ->
     erlang:nif_error(undefined).
+
+% Patch reason: mock implementation, prevents elixir_module:format_error/1 from crashing
+which(_Module) ->
+    non_existing.
+
+% Patch reason: mock implmenetation, prevents ParallelChecker.cache_module/2 from crashing
```
jgonet:
```suggestion
% Patch reason: mock implementation, prevents ParallelChecker.cache_module/2 from crashing
```

fission_lib/test/support/atomvm.ex
```diff
@@ -69,6 +69,24 @@ defmodule FissionLib.Support.AtomVM do
     info.result
   end
 
+  @doc """
+  Evaluates a string using precompiled evaluation code.
+
+  Doesn't crash in case of failure, always returns an `info` map.
+  """
+  def try_eval(code, type, opts \\ []) when is_binary(code) and is_eval_type(type) do
```
jgonet, line 77: We could add `ignore_errors` to opts in `eval` too.

fission_lib/test/fission_lib/elixir_module_test.exs
```diff
@@ -112,4 +112,29 @@ defmodule FissionLib.ElixirModuleTest do
...
+      end
+
+      res2 = W.bar(10, 20, 30)
+      {res1, res2}
+      """
+      |> AtomVM.try_eval(:elixir, run_dir: dir)
+
+    assert %{exit_status: 0, result: {{10, 20}, {20, 30}}} = info
+    logs = File.read!(info.log_path)
+    assert String.contains?(logs, "warning: {unused_var,c,false}")
```
jgonet, line 136: Or `"warning: {unused_var,c,false}" in logs`

mat-hek: `logs` is a binary, I don't think it makes sense to change it to a list

fission_lib/test/support/atomvm.ex
mat-hek: `eval` returns the result term, while `try_eval` returns the `info` map, so it wouldn't be sufficient. Also, it's now compatible with `run` and `try_run`

PR #59, opened 2025-03-12
Fix boolean operators

FranekKubis:
Fixing boolean operators by adding implementation of 'in' macro to patches.

### DIFF
--- a/fission_lib/patches/exavmlib/Kernel.ex
+++ b/fission_lib/patches/exavmlib/Kernel.ex
@@ -210,7 +210,7 @@ defmodule Kernel do
   end
 
 # Patch reason: the "left in right" macro is not capable of 
-# being used in guard is it is defined outside of Kernel.ex patch
+# being used in guard if it is defined outside of Kernel.ex patch
 
   @doc guard: true
   defmacro left in right do


fission_lib/patches/exavmlib/Kernel.ex
mat-hek, line 1: - Add `Patch reason: ` comment(s), explaining why we need this patch
- Do we change the private functions? If not, please remove them and replace calls to them with `foo(...)` -> `:flb_module.foo(...)`

fission_lib/patches/exavmlib/Kernel.ex
```diff
@@ -208,4 +208,54 @@ defmodule Kernel do
   defp struct(%_{} = struct, fields, fun) do
     Enum.reduce(fields, struct, fun)
   end
+
+# Patch reason: the "left in right" macro is not capable of 
+# being used in guard is it is defined outside of Kernel.ex patch
```
mat-hek:
```suggestion
# being used in guard because it is defined outside of Kernel.ex patch
```

That's weird... may be a patching issue 🤔 Can you create an issue on Linear?

FranekKubis: *if - not because
I added an issue to linear.

PR #61, opened 2025-03-13
Disable warnings again

mat-hek:
Since it turns out we can't use Fibonacci strategy (https://github.com/software-mansion-labs/FissionVM/pull/107), which made warnings work (sometimes at least), we need to disable warnings again and try reenabling when we have some progress in fixing GC

Depends on https://github.com/software-mansion-labs/FissionVM/pull/107

### DIFF
--- a/fission_lib/config/config.exs
+++ b/fission_lib/config/config.exs
@@ -2,8 +2,7 @@ import Config
 
 config :fission_lib,
   add_tracing: false,
-  avm_source:
-    {:git, "git@github.com:software-mansion-labs/FissionVM.git", ref: "revert-104-gc-fibonacci"}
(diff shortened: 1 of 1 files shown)


fission_lib/config/config.exs
```diff
@@ -2,7 +2,8 @@ import Config
 
 config :fission_lib,
   add_tracing: false,
-  avm_source: {:git, "git@github.com:software-mansion-labs/FissionVM.git"}
+  avm_source:
+    {:git, "git@github.com:software-mansion-labs/FissionVM.git", ref: "revert-104-gc-fibonacci"}
```
jgonet: Change branch before merging.

PR #66, opened 2025-03-20
Fix string interpolation

FranekKubis:
String interpolation inside the evaluation code (eval_bits:eval_exp_field/6) is using `<<Val:Size/binary-unit:1>>` and I found out that unit option is not working correctly.

This issue will be fixed by patching a function from eval_bits module.

### DIFF
--- a/fission_lib/patches/estdlib/eval_bits.erl
+++ b/fission_lib/patches/estdlib/eval_bits.erl
@@ -1,3 +1,4 @@
+%% Patch reason: in AVM the option "-unit:Size" inside bitstrings is not supported
 -module(eval_bits).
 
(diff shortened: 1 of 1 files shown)


fission_lib/patches/estdlib/eval_bits.erl
```diff
@@ -0,0 +1,116 @@
+-module(eval_bits).
+
+-export([expr_grp/3,expr_grp/4]).
+
+-define(STACKTRACE,
+        element(2, erlang:process_info(self(), current_stacktrace))).
+
+%% Patch reason: in AVM the option "-unit:Size" inside bitstrings is not supported
```
mat-hek: either move this comment on top of the module or copy&paste it before each function it applies to

PR #71, opened 2025-03-25
Add already fixed hexdocs tests

FranekKubis:
Some of the test cases in FissionLib.HexdocsTest have already been fixed by adding Kernel.in to patches. This fix added those test cases to current test scope.

### DIFF
--- a/fission_lib/test/fission_lib/hexdocs_test.exs
+++ b/fission_lib/test/fission_lib/hexdocs_test.exs
@@ -418,8 +418,6 @@ defmodule FissionLib.HexdocsTest do
   # case, cond, and if ===================================================================================================
   # =======================================================================================================================
 
-  #  todo 8 Unused variables causes eval to fail (in the following "_x" works just fine)
-  @tag :skip
   """
   case {1, 2, 3} do
     {1, x, 3} ->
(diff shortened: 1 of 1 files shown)


fission_lib/test/fission_lib/hexdocs_test.exs
```diff
@@ -612,13 +609,13 @@ defmodule FissionLib.HexdocsTest do
...
-  #  :erlang.is_atom(:hello)
-  #  """
-  #  |> assert_eval(true)
+  """
+  fun = &is_atom/1
+  :erlang.is_atom(:hello)
+  """
+  |> assert_eval(true)
 
+  #  todo 12 guards as anonymous functions are not working (adding module to the function [Kernel. or :erlang.] does not help)
```
bblaszkow06, line 628: 1. Do we have issues/sub-issues on board to track those TODOs?
2. Is it possible to skip the test with this line as a reason? I.E.
```suggestion
  @tag skip: "TODO 12: guards as anonymous functions are not working (adding module to the function [Kernel. or :erlang.] does not help)"
```

FranekKubis: 1. Not for all of them - but I don't think they are necessary, we had a talk with Kuba about it while I was writing the tests.
2. It should be possible, Ill check that.

PR #73, opened 2025-03-27
Use new API for hexdocs tests

jgonet:


fission_lib/test/fission_lib/hexdocs_test.exs
```diff
@@ -1,3769 +1,2534 @@
...
+          keys = unsupported |> Keyword.keys() |> inspect()
+          IO.warn("Unsupported options: #{keys}")
+        end
+
+        info =
+          "#{common}\n#{input}"
+          |> Helper.maybe_wrap_in_try(opts)
+          |> AtomVM.try_eval(:elixir, run_dir: dir)
+
+        assert info.exit_status == 0, """
```
FranekKubis, line 31: Here when we have input == "" because it is a single snippet test case maybe we should print it without "Common code:" part.

fission_lib/test/fission_lib/hexdocs_test.exs
```diff
@@ -1,3769 +1,2534 @@
...
+        for logs.
+
+        Snippet:
+        #{input}
+
+        Common code:
+        #{common}
+        """
+
+        message = fn type, expected ->
```
FranekKubis, line 43: Same as line 31.

PR #76, opened 2025-04-04
Patch binary:compile_pattern/1 function.

FranekKubis:


fission_lib/patches/estdlib/binary.erl
```diff
@@ -115,3 +115,11 @@ split(_Binary, _Pattern) ->
 -spec split(Binary :: binary(), Pattern :: binary(), Option :: [global]) -> [binary()].
 split(_Binary, _Pattern, _Option) ->
     erlang:nif_error(undefined).
+
+%% Patch reason: String.split is using patterns compiled by compile_pattern/1
+%% AtomVM does not provide such feature. Luckily all the functions that use compiled pattern
+%% also use just a binary pattern.
+-spec compile_pattern(Pattern) -> binary() when
```
jgonet, line 124: Please mention that compiled tuple is a tuple in OTP and we intentionally do something different.

PR #84, opened 2025-04-11
Ensure implementation of system_info(os_type)

FranekKubis:
Deleted patch. Added test to cover os.type/0 function call.


fission_lib/test/fission_lib/eval_test.exs
```diff
@@ -502,4 +502,14 @@ defmodule FissionLib.EvalTest do
...
     """
   end
+  
+  async_test ":os.type/0", %{tmp_dir: dir} do
+    os_type = :os.type()
+    """
+    :os.type()
+    """
+    |> AtomVM.eval(:elixir, run_dir: dir)
+    |> AtomVM.assert_result(^os_type)
```
jgonet, line 512: While this may work on natively compiled AtomVM, it's not given that it'd work on WASM. Can you check that manually?

FranekKubis: checked

PR #85, opened 2025-04-11
Ensure proper implementation of filename:split/1

FranekKubis:
Deleted patch. Added test to cover filename:split/1 function call.

### DIFF
--- a/fission_lib/test/fission_lib/eval_test.exs
+++ b/fission_lib/test/fission_lib/eval_test.exs
@@ -545,4 +545,12 @@ defmodule FissionLib.EvalTest do
     |> AtomVM.assert_result(^os_type)
   end
   
+
+  async_test ":filename.split/1", %{tmp_dir: dir} do
+    """
+    :filename.split("path/to/a/file")
+    """
+    |> AtomVM.eval(:elixir, run_dir: dir)
+    |> AtomVM.assert_result(["path", "to", "a", "file"])
+  end
 end
--- a/fission_lib/patches/estdlib/filename.erl
+++ b/fission_lib/patches/estdlib/filename.erl
@@ -1,18 +1,5 @@
 -module(filename).
--compile({flb_patch_private, unix_splitb/1}).
 -compile({flb_patch_private, filename_string_to_binary/1}).
-
-unix_splitb(Name) ->
-    %Patch reason: binary:split using list of separators.
-    L = binary:split(Name, <<"/">>, [global]),
-    LL =
-        case L of
-            [<<>> | Rest] when Rest =/= [] ->
-                [<<"/">> | Rest];
-            _ ->
-                L
-        end,
-    [X || X <- LL, X =/= <<>>].
 
 filename_string_to_binary(List) ->
     %Patch reason: changed unicode to utf8 due to lack of support.


fission_lib/test/fission_lib/eval_test.exs
bblaszkow06, line 1: This file gets bigger and bigger. IMO it's the right moment to start thinking about splitting it

fission_lib/test/fission_lib/eval_test.exs
```diff
@@ -502,4 +502,12 @@ defmodule FissionLib.EvalTest do
     end
     """
   end
+
+  async_test ":filename.split/1", %{tmp_dir: dir} do
+    """
+    :filename.split("path/to/a/file")
+    """
```
bblaszkow06, line 552: I really don't like writing those strings, even if they are short. I'd suggest adding a macro that would convert the code to string.
This way we could write it as code, having LSP, formatting, warnings etc.
I did something like that in https://github.com/software-mansion-labs/elixir-wasm/blob/ac18c9bfcaaccb9b0219c7b2dfd0b9680f2a56b3/fission_lib/test/fission_lib/patch_test.exs but the tests haven't made it to the PR

Maybe I could add it in a separate PR? WDYT? @mat-hek @jakub-gonet

fission_lib/test/fission_lib/eval_test.exs
mat-hek: @jakub-gonet merged things into it not that long ago :D

FranekKubis: It would be good to engage @jakub-gonet in this discussions. Last time we had multiple files responsible for testing we went the merging way instead of staying split.
Also there should be a clear separation by the logic of the evaluation test cases and I do not really see such separation possibility.

fission_lib/test/fission_lib/eval_test.exs
mat-hek: Agreed, it used to be like that, not sure why it changed. For erlang code you still need strings though

mat-hek: Just `Macro.to_string` should do the trick

FranekKubis: I do not have strong opinion here. String approach was good to see clearly that the code will be loaded "somewhere as a string".
But as long as the code stays in `do:` block I think it is readable and clear also, so LGTM overall.

fission_lib/test/fission_lib/eval_test.exs
jgonet: We merged it to one since there was no clear separation between them and other files were small anyways. IMO we should have every eval test in one file but improve readability by using `describe` for sections and compress similar code to some for-comprehensions (e.g. exceptions tests take a lot of space in the file and aren't that interesting).

fission_lib/test/fission_lib/eval_test.exs
jgonet: My only problem with that is that you don't see it as "foreign" code that can't use local variables, etc. It's not a big deal though. I'd think of a way that we can pack a lot of tests without much boilerplate but still ensure that they're run simultaneously (linked `patch_test.exs` do it in one `async_test` sequentially).

mat-hek: If we had the code in a quote block explicitly:

```elixir
quote do
  :filename.split("path/to/a/file")
end
|> AtomVM.eval(...)
```

then it would be quite clear that it's not part of the test code IMO

PR #88, opened 2025-04-14
Use OTP supervisor with hibernate patch

bblaszkow06:
This allows starting supervisor with `:one_for_all` strategy

cc: @FKubisSWM @jakub-gonet

### DIFF
--- a/wiki.md
+++ b/wiki.md
@@ -1,15 +1,18 @@
 # Wiedza powszechna o Elixir WASM
 
 ### Jak patchować funkcje w fission_lib?
+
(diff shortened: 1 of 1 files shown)


fission_lib/patches/estdlib/supervisor.erl
```diff
@@ -18,78 +18,186 @@
 % SPDX-License-Identifier: Apache-2.0 OR LGPL-2.1-or-later
 %
 
+
+% Why is this patch needed?
```
FranekKubis:
```suggestion
% Patch reason:
```

jgonet: Use "Patch reason" instead please.

PR #92, opened 2025-04-17
Remove code:get_object_code/1 patch

FKubisSWM:
Removing of the patch waits for: https://github.com/software-mansion-labs/FissionVM/pull/126


mat-hek:
I believe it still needs to wait due to https://github.com/software-mansion-labs/FissionVM/pull/139 :(

mat-hek (approved):

PR #95, opened 2025-04-23
Enable protocols tests, remove protocols patches

mat-hek:
Protocols seem to work well, and patches aren't needed. There's one thing, though: the tests that involve the creation and implementation of the protocols take a very long time (a few minutes) to run. Not sure why - maybe the protocols are re-consolidated? I don't think it's worth diving into it now, but we can ask Jose

cc @jakub-gonet @bblaszkow06

### DIFF
--- a/fission_lib/test/fission_lib/hexdocs_test.exs
+++ b/fission_lib/test/fission_lib/hexdocs_test.exs
@@ -1507,7 +1507,11 @@ defmodule FissionLib.HexdocsTest do
           output: 0
         }
       ]
-    },
+    }
+  ]
+  |> create_tests(category: :protocols, tag: :long_running, tag: [timeout: :timer.minutes(15)])
+
+  [
     {"Enum.reduce(1..3, 0, fn x, acc -> x + acc end)", output: 6},
     {~s|"age: \#{25}"|, output: "age: 25"},
     {
@@ -1526,7 +1530,7 @@ defmodule FissionLib.HexdocsTest do
     },
     {"inspect &(&1+2)", predicate: &(&1 =~ "#Function"), skip: true}
   ]
-  |> create_tests(category: :protocols, tag: [timeout: :timer.minutes(15)])
+  |> create_tests(category: :protocols)
 
   # Comprehensions
   [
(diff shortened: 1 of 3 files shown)


fission_lib/test/fission_lib/hexdocs_test.exs
```diff
@@ -4,13 +4,18 @@ defmodule FissionLib.HexdocsTestHelper do
...
       @tag skip: Keyword.get(unquote(opts), :skip, false)
-      @tag unquote(tag)
+      for tag <- unquote(tags) do
+        @tag tag
+      end
+
       async_test Helper.test_name(
                    unquote(common) <> unquote(input),
-                   unquote(tag)
+                   unquote(hd(tags))
```
jgonet: What is first tag typically?

fission_lib/test/support/atomvm.ex
```diff
@@ -134,7 +134,8 @@ defmodule FissionLib.Support.AtomVM do
       exit_status: exit_status,
       output: output,
       result: result,
-      log_path: log_path
+      log_path: log_path,
+      lldb: "\nAVM_RUN_DIR='#{run_dir}' lldb '#{@atomvm_path}' '#{bundle_path}'"
```
jgonet: I don't like storing this, can we show it only when test fails? We should be able to do that if we expose atom vm path.

fission_lib/test/fission_lib/hexdocs_test.exs
mat-hek: The 'category', like `:protocols`, `:comprehensions` etc. I only added the possibility to add more tags

fission_lib/test/fission_lib/hexdocs_test.exs
```diff
@@ -4,8 +4,8 @@ defmodule FissionLib.HexdocsTestHelper do
   alias __MODULE__, as: Helper
   import FissionLib.Support.AsyncTest
 
-  defmacro test_ast(common, input, opts, tags) do
-    tags = List.wrap(tags)
+  defmacro test_ast(common, input, opts, category, tags) do
+    tags = [category] ++ List.wrap(tags)
```
jgonet, line 8: Nit: `[category | List.wrap(tags)]`

PR #97, opened 2025-04-24
Add JS library

jgonet:
Main API:
```js
const fission = await Fission.init({
     bundleName: "app.avm",
     debug: true
});

// fission.call() will have two args in the future. We won't separete action and args.
try {
    const { data, durationMs } = await fission.call(action, code, { timeoutMs: 10_000 });
} catch (error) {
    // ...
}
```

### DIFF
--- a/iex_wasm/.gitignore
+++ b/iex_wasm/.gitignore
@@ -30,3 +30,6 @@ iex_wasm-*.tar
 
 # secret config
 config.secret.exs
+
+# wasm build dir
+static/wasm
--- a/iex_wasm/static/.gitignore
+++ b/iex_wasm/static/.gitignore
@@ -1,2 +1,0 @@
-/wasm
-/app.avm


iex_wasm/lib/app.ex
```diff
@@ -1,35 +1,36 @@
...
     Process.register(self(), :main)
-    loop()
+    {:ok, nil}
   end
 
-  defp loop() do
-    receive do
-      {:emscripten, {:call, promise, message}} ->
-        {type, code} = type(message)
+  def handle_js_call({action, code}, _promise, state) do
```
jgonet: Elixir API in `App.Fission.JsServer` will likely change. We'll remove callbacks since you can't add more clauses for `handle_info` defined in `defmacro __using__`. They're needed to handle other messages like click handlers.

iex_wasm/lib/fission.ex
```diff
@@ -0,0 +1,87 @@
...
+  @doc false
+  def deserialize(message) do
+    message
+    |> String.split(":", parts: 2)
+    |> List.to_tuple()
+  end
+
+  @doc false
+  def serialize(term) do
+    :io_lib.format(~c"~p", [term])
```
mat-hek: - I think it now doesn't work with strings, let's change that to `inspect`
- It works for iex_wasm, but for any other use case we need some actual serialization, to be able to interpret returned values on the JS side

iex_wasm/lib/fission.ex
```diff
@@ -0,0 +1,87 @@
...
+          {:stop, reason, reply, new_state} ->
+            JsServer.reject(reply, promise)
+            {:stop, reason, new_state}
+
+          {:stop, reason, new_state} ->
+            {:stop, reason, new_state}
+        end
+      end
+
+      @doc false
```
mat-hek:
```suggestion
      @impl true
```

iex_wasm/lib/fission.ex
```diff
@@ -0,0 +1,87 @@
...
+      def handle_js_call(msg, _promise, state) do
+        {:stop, :not_implemented, state}
+      end
+
+      @doc false
```
mat-hek:
```suggestion
      @impl true
```

iex_wasm/static/fission.js
```diff
@@ -0,0 +1,270 @@
...
+  _listenerRef = null;
+
+  _iframe = null;
+  _mountPromise = null;
+  _heartbeatTimeout = null;
+
+  constructor(params, token) {
+    if (token !== INIT_TOKEN) {
+      throw new Error(
+        "Don't construct object directly, use Fission.init() instead",
```
mat-hek:
```suggestion
        "Don't construct the Fission object directly, use Fission.init() instead",
```

iex_wasm/static/fission.js
```diff
@@ -0,0 +1,270 @@
...
+    this.heartbeatTimeoutMs = heartbeatTimeoutMs ?? HEARTBEAT_TIMEOUT_MS;
+  }
+
+  get iframe() {
+    return this._iframe;
+  }
+
+  static async init({ container, ...constructorParams }) {
+    const fission = new Fission(constructorParams, INIT_TOKEN);
+    await fission._mount(container);
```
mat-hek: Why would anyone choose a different container than `body` or `html`? I'd at least made one of them the default

iex_wasm/lib/fission.ex
mat-hek, line 1: Why is it in `iex_wasm`, not `fission_lib`?

iex_wasm/lib/app.ex
mat-hek: I think you can, since you didn't mark `handle_info` as overridable

iex_wasm/lib/fission.ex
```diff
@@ -0,0 +1,87 @@
...
+      def handle_js_cast(msg, state) do
```
mat-hek: this isn't called anywhere

iex_wasm/lib/fission.ex
jgonet: Will change for inspect, further serialization is coming shortly.

iex_wasm/static/fission.js
jgonet: I guess sometimes you'd want to make few iframes and mount them at different points. Or if you use react to mount them at some component. Will make `body` a default.

iex_wasm/lib/fission.ex
jgonet: Mostly because API is evolving. After I'm done with Elixir changes I'll cleanup both iex_wasm and my todo app playground.

iex_wasm/lib/fission.ex
jgonet: Yup, it's for completeness, WASM module exposes call and cast.

iex_wasm/static/fission.js
mat-hek: In this case, the iframe is not displaying anything, so I guess that the only important thing is not to accidentally remove it when it's within another element

iex_wasm/lib/fission.ex
mat-hek: But it's not working, is it?

jgonet: AFAIK `cast` works.

mat-hek: It can't work since `handle_js_cast` is never called, can it?

PR #98, opened 2025-04-25
Game of life demo

bblaszkow06:
Closes EXW-199

### DIFF
--- a/game_of_life/lib/game_of_life.ex
+++ b/game_of_life/lib/game_of_life.ex
@@ -9,7 +9,8 @@ defmodule GameOfLife do
   A simple, example simulation with 3x3 grid and 5 generations
   """
   def start() do
-    {:ok, _app} = GameOfLife.Application.start(:normal, [])
+    puts("Starting simulation\n")
+    _ = GameOfLife.Application.start(:normal, [])
 
     {:ok, _sup, %{grid_pid: pid}} =
       GameOfLife.Supervisor.start_simulation(3, 3, [{0, 1}, {1, 1}, {2, 1}])
@@ -20,11 +21,14 @@ defmodule GameOfLife do
     puts("\n")
 
     for _ <- 1..5 do
-      grid = Grid.tick(pid)
-      grid_str = print_grid(grid)
-      puts(grid_str)
-      puts("\n")
+      _ = Grid.tick(pid)
     end
+
+    grid = Grid.tick(pid)
+    grid_str = print_grid(grid)
+    puts(grid_str)
+    puts("\n")
+    puts("Simulation ended\n")
 
     :ok
   end
--- a/game_of_life/lib/game_of_life/cell.ex
+++ b/game_of_life/lib/game_of_life/cell.ex
@@ -147,6 +147,10 @@ defmodule GameOfLife.Cell do
     for dx <- [-1, 0, 1], dy <- [-1, 0, 1], dx != 0 or dy != 0 do
       {x + dx, y + dy}
     end
-    |> Enum.filter(fn {x, y} -> x >= 0 and x < xSize and y >= 0 and y < ySize end)
+    |> Enum.filter(&inside_grid?(&1, xSize, ySize))
+  end
+
+  defp inside_grid?({x, y}, xSize, ySize) do
+    x >= 0 and x < xSize and y >= 0 and y < ySize
   end
 end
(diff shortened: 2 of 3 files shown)


game_of_life/lib/game_of_life.ex
```diff
@@ -0,0 +1,61 @@
...
+  alias GameOfLife.Grid
+
+  @doc """
+  A simple, example simulation with 3x3 grid and 5 generations
+  """
+  def start() do
+    {:ok, _app} = GameOfLife.Application.start(:normal, [])
+
+    {:ok, _sup, %{grid_pid: pid}} =
+      GameOfLife.Supervisor.start_simulation(3, 3, [{0, 1}, {1, 1}, {2, 1}])
```
jgonet, line 16: How does it fare with bigger boards, performance-wise?

game_of_life/lib/game_of_life/cell.ex
```diff
@@ -0,0 +1,152 @@
...
+          x :: non_neg_integer(),
+          y :: non_neg_integer(),
+          xSize :: pos_integer(),
+          ySize :: pos_integer()
+        ) :: [Grid.coords()]
+  def neighbours(x, y, xSize, ySize) do
+    for dx <- [-1, 0, 1], dy <- [-1, 0, 1], dx != 0 or dy != 0 do
+      {x + dx, y + dy}
+    end
+    |> Enum.filter(fn {x, y} -> x >= 0 and x < xSize and y >= 0 and y < ySize end)
```
jgonet: Nit: can we extract predicate to have it named? `inside_grid?` or something.

game_of_life/lib/game_of_life/simulation.ex
```diff
@@ -0,0 +1,114 @@
...
+  """
+  @spec cell_via_tuple(term(), Grid.coords()) :: GenServer.name()
+  def cell_via_tuple(game_id, coords) do
+    {:via, Registry, {@registry_name, {game_id, coords}}}
+  end
+
+  @doc """
+  Obtains a pid of `GameOfLife.Cell` with provided coords
+  """
+  @spec cell_pid(term(), Grid.coords()) :: pid()
```
jgonet:
```suggestion
  @spec cell_pid(term(), Grid.coords()) :: pid() | nil
```

game_of_life/lib/game_of_life/simulation.ex
```diff
@@ -0,0 +1,114 @@
...
+  """
+  @spec grid_via_tuple(term()) :: GenServer.name()
+  def grid_via_tuple(game_id) do
+    {:via, Registry, {@registry_name, game_id}}
+  end
+
+  @doc """
+  Obtains a pid of `GameOfLife.Grid` for provided `game_id`
+  """
+  @spec grid_pid(term()) :: pid()
```
jgonet: Ditto.

game_of_life/wasm/AtomVM.wasm
```diff
@@ -0,0 +1 @@
+../../../FissionVM/src/platforms/emscripten/build/src/AtomVM.wasm
```
jgonet, line 1: We probably don't want to symlink here in the future.

game_of_life/wasm/index.html
```diff
@@ -0,0 +1,18 @@
...
+    <title>GameOfLife AVM Demo</title>
+  </head>
+  <body>
+    <h1>Game of Life demo</h1>
+    <p>Please check the console for game_of_life output.</p>
+    <script>
+      // Arguments are loaded using fetch API.
+      var Module = {
+        arguments: ["bundle/bundle.avm"],
+      };
```
jgonet, line 14: This will break in the future, we'll use only ES6 modules.

game_of_life/wasm/server.exs
jgonet, line 1: As discussed, in the future we should provide it from Popcorn and it should be the same across apps.

game_of_life/lib/game_of_life.ex
bblaszkow06: On 30x30 grid:
* Elixir <2sec
* Debug AtomVM - GenServer.call timeout or 1m10sec
* Release build with DEBUG_GC=off 8sec
* WASM ~16-18 secs

Interestingly, starting the simulation and printing the status takes the majority of that time on both WASM and local AtomVM

mat-hek: I'd set the default GC strategy to Fibonacci in FissionVM (again :P), it should bring big speedup (DEBUG_GC will keep the same strategy). The startup time is very interesting indeed

bblaszkow06: https://linear.app/swmansion/issue/EXW-271/investigate-long-simulation-start-on-atomvm

PR #101, opened 2025-05-05
Expand Elixir library

jgonet:
Adds:
- `Wasm.run_js/2`
- `Wasm.send_elixir_ready/1`
- `Wasm.register_event_listener/2`
- `Wasm.unregister_event_listener/1`

Swaps serialization to JSON and changes JS API to take `args` instead of `{action, args}`.
Elixir now notifies JS after init, replacing hard-coded timeout in JS.


fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+      {:ok, %RemoteObject{ref: ref}}
+    end
+  end
+
+  defp prepare_js_result(remote_object, types) do
+    type_with_value = fn type, map ->
+      case js_result(remote_object, type) do
+        {:ok, result} -> {:cont, Map.put(map, type, result)}
+        {:error, error} -> {:halt, {:error, error}}
+      end
```
jgonet: We're returning a map with separate `%{ref: ref, value: value}` keys instead of making `value` part of `RemoteObject` to decouple their lifecycles – if user doesn't need `ref`, it will be garbage collected and removed from JS right away.
I considered making result `{:ok, ref} | {:ok, value}` but there may be instances where you need the value and don't want to pay serialization costs. The map is the most forward compatible IMO. We could also do `{:ok, ref} | {:ok, value}` and do `{:ok, ref, value}` later on if needed.

I'm running contest for better name for this function.

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+        try {
+          return (#{js_function})({
+            wasm: Module,
+            bindings: Module.deserialize('#{serialized_bindings}'),
+            window: window.parent,
+            key: key
+          });
+        } catch (e) {
+          console.error(e);
+        }
```
jgonet, line 259: I'm thinking of adding an `onJSError` option to not hard-code `console.error`. For now it's alright to leave it in, I need to run more tests what happens if passed JS is incorrect.
Note to self: ensure that AtomVM doesn't have any try-catches when running JS.

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+    with {:ok, serialized_bindings} <- serialize(bindings) do
+      code = """
+      (Module, key) => {
+        try {
+          return (#{js_function})({
+            wasm: Module,
+            bindings: Module.deserialize('#{serialized_bindings}'),
+            window: window.parent,
+            key: key
+          });
```
jgonet: Only `bindings` and `window` will be documented. The rest will be exposed but "use at your own risk" kind of thing.

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
   end
 
+  @doc """
+  Registers event listener for element with given `selector`. Events will be sent to the process registered under `target` name.
+  To get event data, specify needed keys in `event_keys` list.
+
+  To unregister listener, use returned ref with `unregister_event_listener/1`
+  """
+  @spec register_event_listener(String.t(), register_event_listener_opts()) :: run_js_return()
+  def register_event_listener(event_name, opts) do
```
jgonet: Do you have any other APIs that are non-trivial to implement yourself and would be a good candidates to be included in this module?

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+      }
+      """
+      |> run_js(bindings: %{main: main_process_name})
+
     :ok
   end
 
+  @doc """
+  Registers event listener for element with given `selector`. Events will be sent to the process registered under `target` name.
+  To get event data, specify needed keys in `event_keys` list.
```
mat-hek, line 192: can't we serialize the whole event? at least by default?

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+    end
+  end
+
+  @doc false
   defp deserialize(message) do
-    String.split(message, ":", parts: 2)
+    Jason.decode(message)
   end
 
+  @doc false
```
mat-hek:
```suggestion
```

iex_wasm/lib/Application.ex
```diff
@@ -1,6 +1,10 @@
 defmodule App.Application do
+  alias FissionLib.Wasm
+
   def start do
-    {:ok, _} = GenServer.start_link(App, [])
+    {:ok, pid} = GenServer.start_link(App, [])
+    Process.register(pid, :main)
+    Wasm.send_elixir_ready("main")
```
mat-hek: Maybe these could be `Wasm.register()`?

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+
+  defp prepare_js_result(remote_object, types) do
+    type_with_value = fn type, map ->
+      case js_result(remote_object, type) do
+        {:ok, result} -> {:cont, Map.put(map, type, result)}
+        {:error, error} -> {:halt, {:error, error}}
+      end
+    end
+
+    with %{} = result = Enum.reduce_while(types, %{}, type_with_value) do
```
mat-hek:
```suggestion
    with %{} <- result = Enum.reduce_while(types, %{}, type_with_value) do
```

fission_lib/lib/fission_lib/wasm.ex
mat-hek: How about `value | ref`, value by default, and a function to deserialize `ref -> value`? I don't like returning `ref` by default, as it seems rarely needed and makes it complicated

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
-    :emscripten.promise_reject(promise, serialize(term))
+    with {:ok, raw_message} <- serialize(term) do
+      :emscripten.promise_reject(promise, raw_message)
+      :ok
+    end
+  end
+
+  @doc """
+  Runs JS code in browser's main thread in the WASM iframe context. Takes a JS function as string. The JS function takes an object with following keys:
+  - `wasm`: the underlying WASM module generated by emscripten.
```
mat-hek: Mention its purpose, maybe point to JS docs somehow?

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+  - `bindings`: the arguments passed from Elixir and deserialized at JS side.
```
mat-hek: Maybe `args` would be a better name?

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
+  - `window`: the main window context. Use it instead of calling (implicitly or explicitly) iframe window in the function.
+  - `key`: a key generated for value returned from function.
```
mat-hek: What's it for?

fission_lib/lib/fission_lib/wasm.ex
```diff
@@ -57,31 +91,190 @@ defmodule FissionLib.Wasm do
...
   def reject(term, promise) do
-    :emscripten.promise_reject(promise, serialize(term))
+    with {:ok, raw_message} <- serialize(term) do
+      :emscripten.promise_reject(promise, raw_message)
+      :ok
+    end
+  end
+
+  @doc """
+  Runs JS code in browser's main thread in the WASM iframe context. Takes a JS function as string. The JS function takes an object with following keys:
```
mat-hek, line 116: some example would be useful

fission_lib/lib/fission_lib/wasm.ex
mat-hek: They all seem documented in `run_js` 🤔

fission_lib/lib/fission_lib/wasm.ex
jgonet: 1. functions in JS implicitly return `undefined` so we need to always store function result.
2. some values can't be deserialized, e.g. JS functions.
3. deserialization have some runtime cost.
4. if we pass ref from elixir, JS need to know how to deserialize it (custom reviver in `JSON.parse`).

Returning ref by default makes sense in this scenario and I don't think we should allow to change serialization just yet. Function can fail so we should return tuple IMO. I'm convinced we shouldn't return a map at this moment.

fission_lib/lib/fission_lib/wasm.ex
jgonet: Needs to be fixed, I documented only those two in landing page docs.

fission_lib/lib/fission_lib/wasm.ex
jgonet: Nope, it contains non-serializable attributes.

iex_wasm/lib/Application.ex
jgonet: We could but I'm afraid it won't convey that it does some serious side-effects. You can't register twice unless you unmount iframe.

fission_lib/lib/fission_lib/wasm.ex
jgonet: We store every value returned from passed function in a JS map. The `RemoteObject` returned by default contains ref that is ref-counted. If number of references drops to zero, it gets deleted from that map. `key` is a number (counter increasing in C impl) that indexes the map.

iex_wasm/lib/Application.ex
mat-hek: Maybe we could have a better name, but register is good enough IMO

fission_lib/lib/fission_lib/wasm.ex
mat-hek: Anyway, wasm may make sense to be returned, as you could use it to send messages back during the function execution

jgonet: The problem here is that we have wasm without any guarantees that timeouts will be handled, typically this is handled at main script and not iframe script.

PR #102, opened 2025-05-06
Add iex_wasm tests and CI

mat-hek:
The tests don't run on the CI yet, as they fail there for unknown reasons.

### DIFF
--- a/iex_wasm/test/browser_test.exs
+++ b/iex_wasm/test/browser_test.exs
@@ -15,7 +15,7 @@ defmodule IexWasm.BrowserTest do
 
     # Wait until the server is ready
     page = Playwright.Browser.new_page(browser)
-    wait_for(fn -> assert %{status: 200} = Playwright.Page.goto(page, url) end)
+    wait_for(fn -> assert %{status: 200} = Playwright.Page.goto(page, url) end, 60_000)
     Playwright.Page.close(page)
 
     [browser: browser, url: url]
@@ -50,7 +50,7 @@ defmodule IexWasm.BrowserTest do
     wait_for(fn -> assert Playwright.Page.text_content(page, ~s|[id="result"]|) == result end)
   end
 
-  defp wait_for(fun, timeout \\ 5000)
+  defp wait_for(fun, timeout \\ 5_000)
 
   defp wait_for(fun, timeout) when timeout <= 0 do
     fun.()
--- a/.github/workflows/iex_wasm_build_test.yml
+++ b/.github/workflows/iex_wasm_build_test.yml
@@ -28,7 +28,6 @@ jobs:
 
     name: IEX WASM build & test
     runs-on: ubuntu-22.04
-    container: emscripten/emsdk
 
     defaults:
       run:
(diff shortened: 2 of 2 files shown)


dockerfile
jgonet, line 1: The file should be named `Dockerfile`.

dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    export MIX_ENV=prod && \
+    rm -rf _build deps static/wasm && mix deps.get && mix deps.compile && \
+    ATOMVM_SOURCE_PATH=/FissionVM mix fission_lib.build_avm --out-dir static/wasm --target wasm && \
+    mix compile && \
+    # Run server script to download and cache its deps
+    echo "\n" | elixir server.exs
+
+# Run tests
+RUN apt-get update && apt-get -y install nodejs && \
+    cd /iex_wasm && export MIX_ENV=test && mix playwright.install && mix test
```
jgonet: Why we're running tests both in github actions and here?

iex_wasm/config/config.exs
```diff
@@ -9,15 +9,19 @@ end
...
 config :fission_lib,
   out_path: "static/wasm/app.avm",
   start_module: App.Application,
   add_tracing: include_tracing,
-  avm_source: {:git, "git@github.com:software-mansion-labs/FissionVM.git"}
+  avm_source:
+    (case System.get_env("ATOMVM_SOURCE_PATH") do
+       nil -> {:git, "git@github.com:software-mansion-labs/FissionVM.git"}
+       path -> {:path, path}
+     end)
```
jgonet: Shouldn't we just default to it inside FissionLib?

iex_wasm/test/browser_test.exs
```diff
@@ -0,0 +1,73 @@
...
+    url = "http://localhost:#{port}"
+
+    Task.start_link(fn -> System.shell("elixir server.exs --port #{port}") end)
+    page = Playwright.Browser.new_page(browser)
+
+    Stream.repeatedly(fn ->
+      Process.sleep(200)
+      Playwright.Page.goto(page, url)
+    end)
+    |> Enum.find(&match?(%{status: 200}, &1))
```
jgonet: This is non-obvious what it does IMO.
Maybe something like that? `navigate(page, url, max_retries)`

iex_wasm/test/browser_test.exs
```diff
@@ -0,0 +1,73 @@
...
+  defp wait_for(fun, timeout) do
+    try do
+      fun.()
+    rescue
+      ExUnit.AssertionError ->
+        dt = 100
+        Process.sleep(dt)
+        wait_for(fun, timeout - dt)
+    end
+  end
```
jgonet: Would be nice if we could react to change with timeout instead of polling.

mat-hek: I tried to achieve that, but it seems that the Elixir's playwright wrapper doesn't expose the needed Playwright API (I might be wrong). However, seems like Playwright also does pooling under the hood.

dockerfile
mat-hek: It's only to validate that you can run the tests in this docker. I removed `mix test` from GH action because it still doesn't work there. As far as I understand, ultimately the GH action is going to use the docker anyway, but testing probably should happen in the run step, not the build step

PR #103, opened 2025-05-06
Add landing page

jgonet:
Initial version, without hosted examples and number of 
- Divide sections to separate pages
- Create a command that copies built artifacts
- Replace all fission uses with popcorn
- Abstract start/0 and default receiver registration
- Write "Limitations" section

<details>
<summary><b>Screenshots</b></summary>

image
image
image
</details>


landing-page/style.css
mat-hek, line 1: Like the design a lot 👌

landing-page/server.exs
mat-hek, line 1: Why do we need it here?

landing-page/index.html
```diff
@@ -0,0 +1,525 @@
...
+                        class="preview-img"
+                    />
+                    <figcaption>
+                        <a href="/game-of-life">Game of life</a>, representing
+                        every cell as a process.
+                    </figcaption>
+                </figure>
+            </section>
+
+            <h2>Getting started</h2>
```
mat-hek, line 77: I think the getting started and especially the API section should be in the docs (and partially in the readme that would also be a part of the docs). Having it here, we need to maintain it in two places, and the landing page isn't the place where I would expect it to be anyway. So I'd just redirect to the readme.
Also, I find it a bit hard to read (perceive the text), maybe because of the font? Or because there's no border/boundary around the text? Anyway, it's personal and not applicable if we move it to the readme.

jgonet: I plan to separate every section to docs, I'm keeping it for now since it's unfinished and it's easier to find redundancies that way.

I'm wondering if we should have docs in the site or in the Github docs, I'm personally not a fan of very long readmes.

landing-page/server.exs
jgonet: It will be used to serve examples, right now we have only html.

landing-page/index.html
mat-hek: by docs I mean hexdocs

PR #105, opened 2025-05-12
Start module generation

bblaszkow06:


patches/estdlib/ets.erl
```diff
@@ -101,3 +102,8 @@ lookup_element(_Table, _Key, _Pos) ->
 -spec delete(Table :: table(), Key :: term()) -> true.
 delete(_Table, _Key) ->
     erlang:nif_error(undefined).
+
+%% Mock implementation, not checking for existance
```
mat-hek:
```suggestion
%% Patch reason: Mock implementation, not checking for existance
```

patches/estdlib/supervisor.erl
```diff
@@ -60,3 +61,25 @@ init_children(State, StartSpec) ->
...
   end.
+
+% Patch reason: monotonic_time doesn't handle integer time units
+add_restart(State) ->
+  I = State#state.intensity,
+  P = State#state.period,
+  R = State#state.restarts,
+  % Replaced code
+  % Now = erlang:monotonic_time(1),
+  Now = erlang:monotonic_time(second),
```
mat-hek, line 72: Can we patch monotonic_time itself?

lib/mix/tasks/popcorn.cook.ex
```diff
@@ -19,18 +19,105 @@ defmodule Mix.Tasks.Popcorn.Cook do
...
+      # Logger fails to start due to missing :persistent_term.get/0 implementation
+      raise "Application :logger not supported!"
+    end
+
+    if :popcorn in apps do
+      raise """
+        :popcorn application should not be part of the deps.
+        Ensure you have `runtime: false` set in deps/1 in your Mix.Project
+      """
+    end
```
mat-hek: Why is that?

patches/estdlib/application_master.erl
```diff
@@ -0,0 +1,36 @@
...
+
+-compile({popcorn_patch_private, relay_to_group_leader/2}).
+
+-record(state, {child, appl_data, children = [], procs = 0, gleader, req = []}).
+
+% Patch reason: There's no user nor init processes (kernel app should start it),
+% so instead of forwarding, handle the IO request here, preventing inifnite loop
+% of relay to self
+relay_to_group_leader(IoReq, State) ->
+  erlang:display(IoReq),
```
FranekKubis: Is erlang:display necessary here?

patches/estdlib/supervisor.erl
bblaszkow06: Nope, it's a NIF so the patch won't work. However, the change (in FissionVM) seems trivial, but I didn't want to delay this PR

lib/mix/tasks/popcorn.cook.ex
bblaszkow06: Because of the deps you added to Popcorn - https://github.com/software-mansion/popcorn/blob/f9dd853e4aa66fa8b9287ab81efb39b792171464/mix.exs#L47
Without `runtime: false`, those deps would become deps of the user's app, attempting to start them as well. This is a bit of workaround, but as long as `popcorn` doesn't have any application/supervisor, this is fine.

The permanent solution would be something we spoke about - separating popcorn tools/patching and runtime into different OTP applications

bblaszkow06: PS. I forgot to mention that these apps were failing to start, even after logger removal
PS.2 I tried removing them and "ensure started" from tasks, but it wasn't straightforward and I dropped this attempt

patches/estdlib/application_master.erl
bblaszkow06: Good catch, it's a debug leftover

lib/mix/tasks/popcorn.cook.ex
mat-hek: What if we removed them from extra_applications and started them manually?

mat-hek: Couldn't we override the `applications` key in popcorn's application spec?

examples/hello_popcorn/mix.exs
```diff
@@ -14,15 +14,16 @@ defmodule HelloPopcorn.MixProject do
...
+      mod: {HelloPopcorn.Application, []}
     ]
   end
 
   # Run "mix help deps" to learn about dependencies.
   defp deps do
     [
       # {:popcorn, github: "software-mansion/popcorn"}
-      {:popcorn, path: "../.."},
+      {:popcorn, path: "../..", runtime: false},
```
mat-hek:
```suggestion
      {:popcorn, path: "../.."},
```

Apply in all examples

README.md
```diff
@@ -20,35 +20,73 @@ The examples are hosted at [popcorn.swmansion.com](https://popcorn.swmansion.com
...
 
-defmodule MyApp.Start do
-  def start() do
-    Popcorn.Wasm.register("main")
-    IO.puts("Hello from WASM")
+  @impl true
+  def init(_init_arg) do
+    Popcorn.Wasm.register(@process_name)
+    IO.puts("Hello from WASM!")
+    :ignore
```
mat-hek:
```suggestion
    state = %{}
    {:ok, state}
```

Maybe this is better since we expect people to build on top of it?

examples/game_of_life/lib/game_of_life.ex
mat-hek, line 1: - Moduledoc is outdated
- Do we need this module? Couldn't `GameOfLife.Ui` register itself in `Popcorn.Wasm`? Otherwise, maybe we could have an option in popcorn.js not to require any registration?

lib/mix/tasks/popcorn.cook.ex
```diff
@@ -36,14 +35,51 @@ defmodule Mix.Tasks.Popcorn.Cook do
...
+        spec =
+          Keyword.update(spec, :applications, [], &Enum.reject(&1, fn x -> x == :logger end))
+
+        {app, spec}
+      end)
+
+    if all_specs != specs do
+      env = __ENV__
+
+      # Intentinally omit line number and whole stacktrace plus ensure file is relative to popcorn root
```
mat-hek:
```suggestion
      # Intentionally omit line number and whole stacktrace, plus ensure file is relative to popcorn root
```

lib/popcorn.ex
```diff
@@ -35,7 +34,7 @@ defmodule Popcorn do
 
     default_options = [
       out_dir: Popcorn.Config.get(:out_dir),
-      start_module: Popcorn.Config.get(:start_module),
+      start_module: nil,
```
mat-hek: leftover?

lib/popcorn.ex
```diff
@@ -12,12 +12,11 @@ defmodule Popcorn do
...
   Generates static artifacts to run the project in the browser.
 
-  `out_dir` and `start_module` are mandatory, unless provided
-  via `config.exs`, for example `config :popcorn, out_dir: "static/wasm, start_module: Start"`
+  `out_dir` and `start_module` are mandatory.
 
   Options:
   - `out_dir` - the directory to write artifacts to
-  - `start_module` - a module with `start/0` function, the application's entry point
+  - `start_module` - a module with `start/0` function, used by AtomVM as starting point
```
mat-hek, line 26: I'm not sure we should keep this option documented. If we do, let's at least describe what happens by default

lib/popcorn/config.ex
```diff
@@ -1,13 +1,13 @@
 defmodule Popcorn.Config do
   @moduledoc false
   @defaults %{
-    start_module: nil,
```
mat-hek, line 4: Why keep this in `Popcorn.cook` but remove from the config?

mat-hek (changes requested):

examples/eval_in_wasm/lib/eval_in_wasm.ex
mat-hek, line 1: this clause

```elixir
    unless Process.whereis(:elixir_config) do
      :elixir.start([], [])
    end
```

in eval/2 is a noop, let's remove it

lib/popcorn.ex
bblaszkow06: Well, that is a bit of a discussion about API - the start module has to be passed to packbeam so it cannot be removed completely, and the tests are using `cook` function without `popcorn.cook` task. IMO `Popcorn.cook` is more of an internal or low-level API and the task is intended for general public, so this is OK. I don't see a reason to not document it.

I decided to keep that option for the cook function not to break anything, but remove the default value taken from config (and config option itself)

mat-hek: Ok, then let's mention that it doesn't start applications

mat-hek: BTW, I just noticed that the start module is only generated when you use a task. IMO it should be also generated when you call `Popcorn.cook` and not pass the start module.

examples/game_of_life/lib/game_of_life.ex
bblaszkow06: > Do we need this module? Couldn't GameOfLife.Ui register itself in Popcorn.Wasm?

There is no process explicitly registering itself; that's why a non-existent "noop" is registered only to signal the end of initialisation. A separate module in Application children seemed like the best way to run something after all other processes of the Application start. I didn't want to hide that code inside of the UI

> Otherwise, maybe we could have an option in popcorn.js not to require any registration?

That would be nice, but I feel like out of the scope of this PR. It could be a part of the generated module if `Wasm` had a different way to signal "ready".

examples/eval_in_wasm/mix.exs
```diff
@@ -17,14 +17,15 @@ defmodule EvalInWasm.MixProject do
...
+      extra_applications: [],
+      mod: {EvalInWasm.Application, []}
     ]
   end
 
   defp deps do
     [
-      {:popcorn, path: "../../"},
-      {:playwright, github: "membraneframework-labs/playwright-elixir", only: :test}
+      {:popcorn, path: "../../", runtime: false},
```
mat-hek: is this necessary?

lib/popcorn.ex
```diff
@@ -164,4 +172,133 @@ defmodule Popcorn do
...
+  defp gather_app_specs([], specs), do: specs
+
+  defp gather_app_specs(apps, specs) do
+    new_apps = Enum.reject(apps, &Map.has_key?(specs, &1))
+
+    new_specs =
+      for app <- new_apps,
+          spec = Application.spec(app),
+          # Ignore missing optional apps
+          spec != nil,
```
mat-hek: this check is redundant, maybe let's write it like
```suggestion
          # Ignore missing optional apps
          (spec = Application.spec(app)) != nil,
```

lib/popcorn.ex
```diff
@@ -42,25 +42,32 @@ defmodule Popcorn do
   end
 
   def bundle(options \\ []) do
-    all_beams = Path.wildcard(Path.join([@app_build_root, "**", "*.{beam,app}"]))
+    {module, filename} = Popcorn.create_boot_module(options[:start_module])
```
mat-hek:
```suggestion
    {module, filename} = create_boot_module(options[:start_module])
```

lib/popcorn.ex
```diff
@@ -164,4 +172,133 @@ defmodule Popcorn do
       """
     end
   end
+
+  def create_boot_module(start_module \\ nil) do
```
mat-hek:
```suggestion
  defp create_boot_module(start_module) do
```

mat-hek (approved):

lib/popcorn.ex
bblaszkow06: It isn't redundant, when app is optional, it appears in applications but spec is missing.

From [app docs](https://www.erlang.org/doc/apps/kernel/app)
> `applications` - All applications that must be started before this application. If an application is also listed in optional_applications, then the application is not required to exist (but if it exists, it is also guaranteed to be started before this one).

The suggested solution is IMHO less readable, one line of code less isn't worth it.

mat-hek: It is redundant, because it will never be executed, because if `spec` is `nil`, the whole assignment will return `nil` and `for` will jump to the next app. So it's not about saving one line, but I agree that my solution is less readable, so maybe something like:

```suggestion
          spec <- [Application.spec(app)],
          # Ignore missing optional apps
          spec != nil,
```

I don't like it very much either, but it's better than having a redundant check IMO

bblaszkow06: You're right, I forgot about this `for` behaviour. IMO the correct solution is to just leave the comment
```suggestion
          # Missing optional apps (`nil` spec) will be filtered out
          spec = Application.spec(app),
```

PR #110, opened 2025-05-14
Added ex docs dependency and reviewed generated docs

FranekKubis:


lib/popcorn/api/wasm.ex
```diff
@@ -188,6 +197,7 @@ defmodule Popcorn.Wasm do
     value
   end
 
+#  todo - discuss if it notifies why is it called register?
```
FranekKubis: Comment will be deleted after discussing.

mix.exs
```diff
@@ -39,7 +39,8 @@ defmodule Popcorn.MixProject do
   defp deps do
     [
       {:atomvm_packbeam, github: "atomvm/atomvm_packbeam"},
-      {:jason, "~> 1.4"}
+      {:jason, "~> 1.4"},
+      {:ex_doc, "~> 0.34", runtime: false, warn_if_outdated: true},
```
bblaszkow06: It should only be used in `:dev` env
```suggestion
      {:ex_doc, "~> 0.34", only: :dev, runtime: false, warn_if_outdated: true},
```

lib/popcorn/api/wasm.ex
bblaszkow06: I think it's because it also "registers" on the JS side the name under which the main Elixir process (i.e. the one that will receive messages) is registered in BEAM

bblaszkow06 (approved):
* task `mix compile.popcorn` has no docs
* There are a few functions without docs

IMHO we can fix that later

PR #111, opened 2025-05-14
iex wasm implementation

FKubisSWM:


examples/iex_wasm/mix.exs
```diff
@@ -0,0 +1,30 @@
...
+      deps: deps(),
+      aliases: [
+        build_wasm: "popcorn.build_runtime --target wasm --out-dir static/wasm"
+      ]
+    ]
+  end
+
+  def application do
+    [
+      extra_applications: [:logger]
```
bblaszkow06: It's not using logger, I would remove it
```suggestion
      extra_applications: []
```

patches/estdlib/io.erl
```diff
@@ -57,6 +57,14 @@ format(Format, Args) when is_list(Format) andalso is_list(Args) ->
         end,
     put_chars(Msg).
 
+%% Patch reason: fwrite should behave just like format in Popcorn
+fwrite(Format) ->
+    format(Format).
+
+%% Patch reason: fwrite should behave just like format in Popcorn
+fwrite(Format, Args) ->
+    format(Format, Args).
```
bblaszkow06, line 66: Is that something that could land in Atom/Fission instead?

patches/estdlib/net_kernel.erl
bblaszkow06, line 1: Considering the patch reason in `bc_req` - does it work? If that's relevant, `dflag_unicode_io` seems to be a BIF in BEAM

BTW patch reason here is missing

patches/exavmlib/Exception.ex
```diff
@@ -79,7 +79,10 @@ defmodule Exception do
   """
   def message(%module{__exception__: true} = exception) do
     try do
-      module.message(exception)
+#      for some reason the latter is crushing when invoked with FunctionClauseError
```
bblaszkow06:
```suggestion
#      for some reason, the latter is crashing when invoked with FunctionClauseError
```

By "latter" you meant the added code with interpolation? Or did you mean the commented line (the "former")

bblaszkow06: It makes sense to debug & fix it (as a separate issue probably)

patches/exavmlib/server.ex
bblaszkow06, line 1: There's only one patch reason for the whole file, feels like something is missing

patches/estdlib/io.erl
jgonet: This is already implemented in upstream.

examples/iex_wasm/README.md
```diff
@@ -0,0 +1,43 @@
+# IEx Wasm
+
+This project aims to be an Elixir IEx running fully in the browser via WASM via AtomVM.
+
+For now, it's a textarea where you can type Erlang code and run it (fully in browser, via WASM via AtomVM).
```
jgonet: Is it correct?

examples/iex_wasm/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+for the WASM AtomVM to work.
+
+## Deploy
+
+- Build `.avm` file and AtomVM WASM binary in release mode (`MIX_ENV=prod mix compile`).
+- Connect to server via ssh.
+- Copy files into the server (`scp -A -r <SOURCE_DIR> root@<IP>:/var/www/wasm-$(date "+%Y-%m-%d")`).
+- Update symlink to point to new directory (in `/var/www/`: `ln -sf ./wasm-DATE html`)
```
jgonet: This is obsolete.

examples/iex_wasm/static/script.js
```diff
@@ -0,0 +1,50 @@
...
+    const { data, durationMs } = await popcorn.call({"command": "start", "language": LANGUAGE}, {
+      timeoutMs: 10_000,
+    });
+  } catch (error) {
+      displayLog(error, { isError: true }) 
+  }
+  
+  window.terminal.onKey(async key => {
+    let code_bunch = key.key;
+    let keyCode = key.domEvent.keyCode;
```
jgonet: 1. What is code bunch?
2. Use `const`.
3. Use camelCase consistently.

examples/iex_wasm/static/script.js
```diff
@@ -0,0 +1,50 @@
...
+    if (keyCode === 38 || keyCode === 40) {
```
jgonet: Please extract 38 and 40 to named constants.

examples/iex_wasm/static/script.js
```diff
@@ -0,0 +1,50 @@
+import { Popcorn } from "./wasm/popcorn.js";
+
+const LANGUAGE = document.querySelector('meta[name="code-language"]').content;
+
+var term = new Terminal();
+term.open(document.getElementById(LANGUAGE + '_terminal'));
+window.terminal = term;
```
jgonet: 1. Don't use `var`
2. It doesn't make sense to have `setup` function and do init stuff outside of it.

patches/estdlib/io.erl
FranekKubis: Yes, but as I already noticed somewhere - the erlang implementation of some functions in AtomVM is overwritten by the OTP implementation in Popcorn. Either we change the way we overwrite erlang module or we try to make OTP work.
All in all IO should be treated as a one compact issue to be resolved in the future and I would leave this patch for now.

patches/estdlib/net_kernel.erl
FranekKubis: Actually you are right - the patch is not necessary. I removed it.

patches/exavmlib/Exception.ex
FranekKubis: My suggestion:

`for some reason, the 'module.message(exception)' is crashing when invoked with FunctionClauseError`

FranekKubis: Yes, of course.

patches/exavmlib/server.ex
FranekKubis: I clean the file up.

examples/iex_wasm/README.md
FranekKubis: I removed the lines

examples/iex_wasm/static/script.js
```diff
@@ -1,14 +1,15 @@
...
 const LANGUAGE = document.querySelector('meta[name="code-language"]').content;
-
-var term = new Terminal();
-term.open(document.getElementById(LANGUAGE + '_terminal'));
-window.terminal = term;
+const ARROW_UP_KEY_CODE = 38;
+const ARROW_DOWN_KEY_CODE = 40;
 
 async function setup() {
+  var term = new Terminal();
```
bblaszkow06:
```suggestion
  const term = new Terminal();
```

bblaszkow06 (approved):

PR #124, opened 2025-05-21
Fix dockerfile

mat-hek:
This PR overrides the AtomVM downloaded by Popcorn with the locally built one. Seems it [still works](https://popcorn.swmtest.xyz/)

### DIFF
--- a/utils/landing-page/Dockerfile
+++ b/utils/landing-page/Dockerfile
@@ -64,13 +64,13 @@ RUN cd /root/elixir-wasm && \
  RUN echo 'export ERL_AFLAGS="+JMsingle true"' >> /root/.env
  RUN source /root/.env && \
      cd /root/elixir-wasm/examples/eval_in_wasm && \
-     cp -r /root/wasm static && \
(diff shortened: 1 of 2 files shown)


utils/landing-page/Dockerfile
```diff
@@ -64,13 +64,13 @@ RUN cd /root/elixir-wasm && \
...
-     mix deps.get && mix deps.compile && mix popcorn.cook && mix compile
+     mix deps.get && mix popcorn.cook && \
+     cp -r /root/wasm static
 
  RUN source /root/.env && \
      cd /root/elixir-wasm/examples/game_of_life && \
-     cp -r /root/wasm static && \
-     mix deps.get && mix deps.compile && mix popcorn.cook && mix compile
+     mix deps.get && mix popcorn.cook && \
+     cp -r /root/wasm static
```
jgonet, line 73: Note: `mix popcorn.cook` downloads binaries from our site, that's why `cp -r` is after. We need to fix up the flow in the near future.

jgonet (approved):

PR #164, opened 2025-05-23
Prepare to publish on Hex

mat-hek:


lib/popcorn/api/wasm.ex
```diff
@@ -308,9 +309,8 @@ defmodule Popcorn.Wasm do
   end
 
   defp run_js_fn(code) do
-    with {:ok, ref} <- :emscripten.run_remote_object_fn_script(code, main_thread: true) do
-      {:ok, %RemoteObject{ref: ref}}
-    end
+    {:ok, ref} = :emscripten.run_remote_object_fn_script(code, main_thread: true)
+    {:ok, %RemoteObject{ref: ref}}
```
jgonet: Why we change that?

mix.exs
```diff
@@ -30,6 +50,26 @@ defmodule Popcorn.MixProject do
   defp elixirc_paths(:test), do: ["lib", "test/support"]
   defp elixirc_paths(_env), do: ["lib"]
 
+  defp package do
+    [
+      maintainers: ["Software Mansion"],
+      licenses: ["Apache-2.0"],
+      files: ["lib", "priv", "patches", "mix.exs", "README*", "LICENSE*", "src"],
```
jgonet: Why not just README.md and LICENSE?

lib/stub/emscripten.ex
```diff
@@ -0,0 +1,24 @@
+defmodule :emscripten do
+  @moduledoc false
+
+  @spec from_remote_object(reference(), :key | :value) :: term()
```
jgonet: It's not exactly reference, it's binary.

lib/popcorn/api/wasm.ex
mat-hek: Because `run_remote_object_fn_script` never returns anything else. But I should be consistent and return plain `RemoteObject`

mix.exs
mat-hek: Because then we'd only publish readme and license on Hex

jgonet: I mean "README*" -> ""README.md"

lib/popcorn/api/wasm.ex
jgonet: Ah, right. We raise on errors there. `get_remote_object` works in a different way though.

mix.exs
mat-hek: ah ok, don't know, copied it over ¯\\_(ツ)\_/¯

mat-hek: changed

lib/popcorn/api/wasm.ex
mat-hek: Yup, it could be more consistent

jgonet (approved):

PR #168, opened 2025-05-28
Browser tests

### LINKED ISSUE #167: Run Popcorn tests in WASM
Currently, the tests are only run on AtomVM compiled for Unix. We should also run them on WASM to make sure they work everywhere.

mat-hek:
closes #167 

TODO:
- [x] Use `Popcorn.cook` to create WASM artefacts for tests
- [x] Investigate the tests that don't pass (5 right now)
- [x] Allow running Unix tests too
- [x] Fix caching eval modules
- [x] Make ~`runtime_path`~ `runtime` work again
- [x] Add `testing` chapter to readme


.github/workflows/popcorn_build_test.yml
```diff
@@ -39,6 +39,10 @@ jobs:
         with:
           ssh-private-key: |
             ${{ secrets.ACCESS_TO_FISSIONVM }}
+      - name: Setup emsdk
+        uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: 4.0.7
```
bblaszkow06, line 45: We can enable cache here: https://github.com/mymindstorm/setup-emsdk#cache
```suggestion
          version: 4.0.7
          actions-cache-folder: 'emsdk-cache'
```

bblaszkow06, line 45: I'd consider bumping emsdk to 4.0.10

.github/workflows/popcorn_build_test.yml
```diff
@@ -61,5 +65,10 @@ jobs:
         run: |
           export PATH=$PATH:/home/runner/.mix/elixir/1-17/ # for rebar3 to be found
           mix popcorn.build_runtime --target unix --cmake-opts "SANITIZER=OFF DEBUG_ASSERTIONS=ON DEBUG_GC=ON"
+          mix popcorn.build_runtime --target wasm
```
bblaszkow06, line 70: I'd run it as a separate step to track exec time easily

lib/mix/tasks/popcorn.build_runtime.ex
```diff
@@ -59,6 +59,10 @@ defmodule Mix.Tasks.Popcorn.BuildRuntime do
         build_dir = Path.join(runtime_source, "src/platforms/emscripten/build")
         File.mkdir_p!(build_dir)
 
+        unless match?({_path, 0}, System.shell("which emcmake")) do
+          raise "emcmake command not found, please make sure you have emscripten installed"
+        end
```
bblaszkow06: 1.`unless` is deprecated
2. You don't have to use `System.shell`, there's `find_executable`

```suggestion
        if System.find_executable("emcmake") == nil do
          raise "emcmake command not found, please make sure you have emscripten installed"
        end
```

lib/popcorn.ex
```diff
@@ -23,12 +29,15 @@ defmodule Popcorn do
   - `compile_artifacts` - compiled BEAMs and other artifacts that should be included
   in the generated bundle. Defaults to all the `.beam` and `.app` files for the application
   and dependencies.
+  - `bundle_only` - only generates `bundle.avm` file. All other necessary files have to
```
bblaszkow06: Maybe instead of such option it would be best to have separate functions and separate tasks for each "step" + additional (let's say `popcorn.cook`) that calls them all

lib/popcorn.ex
```diff
@@ -37,24 +46,40 @@ defmodule Popcorn do
...
     options = options |> Keyword.validate!(default_options) |> Map.new()
-    ensure_option_present(options, :out_dir, "output directory")
-    ensure_option_present(options, :start_module, "start module")
+    ensure_option_present(options, :out_dir, "Output directory")
+    ensure_option_present(options, :start_module, "Start module")
 
     File.mkdir_p!(options.out_dir)
-    copy_runtime_artifacts(options)
+
+    unless options.bundle_only, do: copy_runtime_artifacts(options)
```
bblaszkow06:
```suggestion
    if not options.bundle_only, do: copy_runtime_artifacts(options)
```

lib/popcorn.ex
```diff
@@ -120,15 +145,24 @@ defmodule Popcorn do
     end
 
     atomvm_artifacts_dir = Path.join([@popcorn_path, "atomvm_artifacts", "#{options.target}"])
+
+    unless File.exists?(atomvm_artifacts_dir) do
```
bblaszkow06:
```suggestion
    if not File.exists?(atomvm_artifacts_dir) do
```

test/popcorn/compilation_test.exs
```diff
@@ -63,10 +63,12 @@ defmodule Popcorn.CompilationTest do
 
     assert [
              {:orddict, :take, 2, [file: ~c"orddict", line: 118]},
-             {RunExpr, :run, 1, [file: code_file, line: run_line]},
-             {RunExpr, :start, 0, [file: code_file, line: start_line]}
+             {RunExpr, :run, 1, [file: code_file, line: run_line]}
+             | _rest
            ] = stacktrace
 
+    assert {RunExpr, :start, 0, [file: ^code_file, line: start_line]} = List.last(stacktrace)
```
bblaszkow06: nit: you could use `rest`
```suggestion
             | rest
           ] = stacktrace

    assert {RunExpr, :start, 0, [file: ^code_file, line: start_line]} = List.last(rest)
```

test/test_helper.exs
```diff
@@ -1,14 +1,24 @@
-alias Popcorn.Support.AtomVM
+alias Popcorn.Support.{AtomVM, Browser}
 
-# FIXME: Workaround for the `Popcorn.Support.AtomVM` caching not considering changing the patches
-File.rm_rf!("tmp")
+Path.wildcard("tmp/*") |> List.delete("tmp/modules") |> Enum.each(&File.rm_rf!/1)
+
+target = System.get_env("TARGET", "UNIX") |> String.downcase() |> String.to_atom()
```
bblaszkow06, line 5: IIUC this won't be run without mix, so you can use `Mix.target()`
```suggestion
target = Mix.target()
```

test/test_helper.exs
```diff
@@ -1,14 +1,24 @@
...
+    Popcorn.ingredients(target: :unix, out_dir: "test/fixtures/unix")
 
-ExUnit.start([capture_log: true, exclude: :long_running] ++ ci_opts)
+  :wasm ->
+    Popcorn.ingredients(target: :wasm, out_dir: "test/fixtures/wasm/static/wasm")
+    Browser.launch()
+end
+
+ci_opts = if System.get_env("CI") == "true", do: [max_cases: 1], else: []
+ExUnit.start([capture_log: true, exclude: [:long_running, :"skip_#{target}"]] ++ ci_opts)
```
bblaszkow06: exclude/include filters accept values, so you should not be using such generated atoms.
Instead, you should be using a tag with value, e.g. `@tag target: :wasm` and then
`exclude: [:target], include: [target: :unix]`

lib/popcorn/utils/fetch_artifacts.ex
```diff
@@ -0,0 +1,78 @@
+defmodule Popcorn.Utils.FetchArtifacts do
+  @moduledoc false
+
+  @artifatcs %{
+    wasm: ["AtomVM.wasm", "AtomVM.mjs"],
+    unix: ["AtomVM"]
+  }
+
+  Application.compile_env(:popcorn, :runtime)
```
bblaszkow06: Result is ignored, do we need that?

lib/popcorn/utils/fetch_artifacts.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  alias Popcorn.Utils.Downloader
+
+  def fetch_artifacts() do
+    runtime_specs =
+      Popcorn.Config.get(:runtime)
+      |> List.wrap()
+      |> Enum.map(fn
```
bblaszkow06, line 15: Why wrap it as a list? AFAIK this will always be one value so you could pipe to `case` or use `then`

lib/popcorn/utils/fetch_artifacts.ex
```diff
@@ -0,0 +1,78 @@
...
+          {type, location, []}
+
+        {type, location, opts}
+        when type in [:path, :url] and is_binary(location) and is_list(opts) ->
+          {type, location, opts}
+      end)
+
+    for {type, location, opts} <- runtime_specs,
+        targets = opts |> Keyword.get(:target, [:unix, :wasm]) |> List.wrap(),
+        target <- targets do
```
bblaszkow06:
```suggestion
        target <- opts |> Keyword.get(:target, [:unix, :wasm]) |> List.wrap() do
```

lib/popcorn/utils/fetch_artifacts.ex
```diff
@@ -0,0 +1,78 @@
...
+      dir = Path.join(Mix.Project.app_path(), "atomvm_artifacts/#{target}")
+      File.mkdir_p!(dir)
+      artifacts = Map.fetch!(@artifatcs, target)
+      paths = Enum.map(artifacts, &Path.join(dir, &1))
+
+      unless Enum.all?(paths, &File.exists?/1) do
```
bblaszkow06:
```suggestion
      if not Enum.all?(paths, &File.exists?/1) do
```

lib/popcorn/utils/fetch_artifacts.ex
```diff
@@ -0,0 +1,78 @@
...
+      end
+    end)
+  end
+
+  defp download_artifact(url, dir, name) do
+    path = Path.join(dir, name)
+
+    with {:ok, _stream} <-
+           Downloader.download("#{url}/#{name}", File.stream!(path <> ".download")) do
+      File.rename!(path <> ".download", path)
```
bblaszkow06: nit:
```suggestion
    tmp_path = path <> ".download"

    with {:ok, _stream} <-
           Downloader.download("#{url}/#{name}", File.stream!(tmp_path)) do
      File.rename!(tmp_path, path)
```

lib/popcorn/utils/fetch_artifacts.ex
```diff
@@ -0,0 +1,78 @@
...
+    Enum.each(artifacts, fn name ->
+      path = Path.join(location, name)
+
+      if File.exists?(path) do
+        File.cp!(path, Path.join(dir, name))
+      else
+        IO.warn("""
+        Couldn't find runtime file #{name} in #{location} \
+        please use mix popcorn.build_runtime to build from source.
+        """)
```
bblaszkow06: Shouldn't that be raised as an error?

test/fixtures/unix/AtomVM
bblaszkow06, line 1: Do we really want that in repo?

test/support/atomvm.ex
```diff
@@ -90,27 +91,28 @@ defmodule Popcorn.Support.AtomVM do
...
     log_path = Path.join(run_dir, "logs.txt")
 
-    args
-    |> Map.new()
-    |> :erlang.term_to_binary()
-    |> then(&File.write(args_path, &1))
+    args |> :erlang.term_to_binary() |> then(&File.write(args_path, &1))
 
-    unless match?({_output, 0}, System.shell("which '#{@runtime_path}'")) do
+    unless match?({_output, 0}, System.shell("which '#{@unix_path}'")) do
```
bblaszkow06: get rid of deprecated unless, however `System.find_executable` won't work with relative path I think

test/support/browser.ex
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def wait_for(fun, timeout) do
+    try do
+      fun.()
+    rescue
+      ExUnit.AssertionError ->
+        dt = 100
+        Process.sleep(dt)
+        wait_for(fun, timeout - dt)
```
bblaszkow06, line 89: this is an infinite loop, there's a clause for `timeout < 0` missing

test/support/browser.ex
```diff
@@ -0,0 +1,86 @@
...
+    """)
+
+    # capture unhandled errors form the iframe window and re-log them again
+    # otherwise Playwright won't capture them
+    Playwright.Page.evaluate(page, """
+    const popcorn_iframe = document.querySelector("iframe")
+    popcorn_iframe.contentWindow.addEventListener("error", (event) => {
+    const message = event.error ? event.error.message : event.message;
+      console.log(`[${event.filename}:${event.lineno}] ${message}`, event)
+    })
```
bblaszkow06:
```suggestion
    const popcorn_iframe = document.querySelector("iframe");
    popcorn_iframe.contentWindow.addEventListener("error", (event) => {
      const message = event.error ? event.error.message : event.message;
      console.log(`[${event.filename}:${event.lineno}] ${message}`, event);
    })
```

test/support/browser.ex
```diff
@@ -0,0 +1,86 @@
...
+  @moduledoc false
+
+  import ExUnit.Assertions
+  import ExUnit.Callbacks
+
+  def launch(options \\ []) do
+    port = Keyword.get(options, :port, 9876)
+    url = "http://localhost:#{port}"
+
+    Task.start_link(fn -> System.shell("elixir test/fixtures/wasm/server.exs --port #{port}") end)
```
bblaszkow06, line 11: Wouldn't it be better to start it as supervised process instead of script in separate process? There would be no need for the `tmp_browser`

config/.credo.exs
```diff
@@ -13,7 +13,8 @@
       checks: %{
         extra: [
           {Credo.Check.Design.TagTODO, [exit_status: 0]},
-          {Credo.Check.Design.TagFIXME, [exit_status: 0]}
+          {Credo.Check.Design.TagFIXME, [exit_status: 0]},
+          {Credo.Check.Readability.WithSingleClause, false}
```
bblaszkow06, line 17: I haven't seen (or maybe missed) any new code violating that rule. If that's the case, we shouldn't add it

bblaszkow06 (changes requested):

.github/workflows/popcorn_build_test.yml
mat-hek: Let's do that in another PR and in all workflows

mat-hek: Let's do that in another PR and in all workflows

config/.credo.exs
mat-hek: I got that error once, so the rule was violated at some point, not sure if it's now, but I'd disable it anyway

lib/popcorn/utils/fetch_artifacts.ex
mat-hek: There can be multiple entries, like

```elixir
config :popcorn,
  runtime: [
    {:path, "/path/to/unix", target: :unix},
    {:path, "/path/to/wasm", target: :wasm}
  ]
```

It needs documentation though :P

test/test_helper.exs
mat-hek: I think targets are more about the architecture the project compiles to, not the architecture to test on. Like we could go for setting MIX_TARGET to WASM and run mix compile instead of popcorn.cook. Here, using MIX_TARGET would likely cause Mix to recompile the entire project when you run `mix test` and then `MIX_TARGET=wasm mix test`, which doesn't make sense.

test/support/browser.ex
mat-hek: This way we also test the server script

mat-hek:
Force pushed due to rebase. The changes since review start with [f612d5b](https://github.com/software-mansion/popcorn/pull/168/commits/f612d5b412045646cffefe12fcb8d17a1b5c5d97)

lib/popcorn/config.ex
```diff
@@ -10,6 +10,7 @@ defmodule Popcorn.Config do
         Path.wildcard("#{Application.app_dir(:iex)}/ebin/**/*.beam"),
     out_dir: nil,
     add_tracing: false,
+    runtime: {:url, "https://popcorn.swmansion.com/simple_repl/wasm/", target: :wasm},
     runtime_source: {:git, "git@github.com:software-mansion-labs/FissionVM.git"}
```
jgonet, line 14: Why we have  both `runtime` and `runtime_source`?

test/test_helper.exs
```diff
@@ -1,14 +1,24 @@
...
+    Popcorn.ingredients(target: :unix, out_dir: "test/fixtures/unix")
 
-ExUnit.start([capture_log: true, exclude: :long_running] ++ ci_opts)
+  :wasm ->
+    Popcorn.ingredients(target: :wasm, out_dir: "test/fixtures/wasm/static/wasm")
+    Browser.launch()
+end
+
+ci_opts = if System.get_env("CI") == "true", do: [max_cases: 1], else: []
+ExUnit.start([capture_log: true, exclude: [:long_running, skip_target: target]] ++ ci_opts)
```
jgonet, line 24: Why CI needs `max_cases: 1`? Also we should consider doing something about long-running tests, excluding them by default ensures nobody will ever run them.

test/popcorn/hexdocs_test.exs
```diff
@@ -145,9 +130,10 @@ defmodule Popcorn.HexdocsTestHelper do
...
   def maybe_wrap_in_try(code, _opts), do: code
 
-  def supported_opts(opts) do
-    supported_opts = [:raises, :output, :predicate, :skip, :cases, :stdout]
-    Keyword.split(opts, supported_opts)
+  def warn_unsupported_opts(opts) do
+    supported_opts = [:category, :tag, :raises, :output, :predicate, :skip, :cases, :stdout]
+    unsupported_opts = opts |> Keyword.drop(supported_opts) |> Keyword.keys()
+    if unsupported_opts != [], do: IO.warn("Unsupported options: #{inspect(unsupported_opts)}")
   end
```
jgonet: Please just raise here. This shouldn't ever happen.

test/fixtures/wasm/static/index.html
```diff
@@ -0,0 +1,11 @@
...
+<head>
+    <script type="module">
+        import { Popcorn } from "./wasm/popcorn.js";
+        window.popcorn_promise = Popcorn.init({
+            bundlePath: new URLSearchParams(window.location.search).get("bundlePath"),
+            onStdout: (output) => console.log(`[popcorn stdout] ${output}`)
+        });
+    </script>
+</head>
+</html>
```
jgonet: Nit: newline at the end.

test/fixtures/unix/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
jgonet: Nit: newline at the end.

lib/popcorn/utils/fetch_artifacts.ex
```diff
@@ -0,0 +1,76 @@
...
+        {type, location} when type in [:path, :url] and is_binary(location) ->
+          {type, location, []}
+
+        {type, location, opts}
+        when type in [:path, :url] and is_binary(location) and is_list(opts) ->
+          {type, location, opts}
+      end)
+
+    for {type, location, opts} <- runtime_specs,
+        target <- opts |> Keyword.get(:target, [:unix, :wasm]) |> List.wrap() do
```
jgonet, line 25: Maybe some variation of that?
```elixir
runtime_specs =
    :runtime
    |> Popcorn.Config.get()
    |> List.wrap()
    |> Enum.map(&parse_spec/1)
    
    defp parse_runtime_config_spec({type, location}) do
        parse_runtime_config_spec({type, location, []})
    end
    
    defp parse_runtime_config_spec({type, location, opts}) do
        true = type in [:path, :url]
        true = is_binary(location)
        
        opts = opts
            |> Keyword.update(:target, [:unix, :wasm], &List.wrap/1)
            |> Keyword.validate!([:target])
            
        {type, location, opts}
    end
```

lib/mix/tasks/popcorn.build_runtime.ex
```diff
@@ -59,6 +59,10 @@ defmodule Mix.Tasks.Popcorn.BuildRuntime do
         build_dir = Path.join(runtime_source, "src/platforms/emscripten/build")
         File.mkdir_p!(build_dir)
 
+        if System.find_executable("emcmake") == nil do
+          raise "emcmake command not found, please make sure you have emscripten installed"
```
jgonet: If we check that, we can also check for `cmake`, `make`:
```elixir
ensure_executables!(["emcmake", "cmake", "make"])

defp ensure_executables!(execs) do
  has_executable? = fn name -> name |> System.find_executable() |> is_binary() end
  for exe <- execs do
    if not has_executable?.(exe) do
      raise "#{exe} command not found, please make sure you have emscripten and C tooling installed"
    end
  end
end
```

lib/popcorn/config.ex
mat-hek: `runtime` is for prebuilt atomvm and `runtime_source` is for building it from source

test/test_helper.exs
mat-hek: I think `max_cases: 1` is because we print logs to stdout instead of saving them to a file, so if we run the tests concurrently, they would be unreadable. And it doesn't make sense to have any significant parallelism given the resources available. I don't know how to go about the long-running tests, since they take minutes to run. Maybe enable them when we speed things up. Neither of these things is related to this PR anyway.

lib/popcorn/utils/fetch_artifacts.ex
mat-hek: I believe extracting it to private functions would only make you jump there and back when reading it

lib/mix/tasks/popcorn.build_runtime.ex
mat-hek: I think having emcmake implies having the rest

lib/popcorn/config.ex
jgonet: I kinda don't like having two keys – if two are specified we need to either error or prioritize one over other.
What do you think about having `runtime: {:git, "<url>", type: :source | :prebuilt}`? This way we have one option and we always have opts which simplifies the code.

mat-hek: Can do, but in a separate PR :P

lib/mix/tasks/popcorn.build_runtime.ex
bblaszkow06: > I think having emcmake implies having the rest

Not necessarily, on my Manjaro, the `emscripten` package does not depend on either cmake or make

.github/workflows/popcorn_build_test.yml
```diff
@@ -57,9 +61,17 @@ jobs:
...
       - name: Lint
         run: mix lint
-      - name: Build AtomVM
+      - name: Build AtomVM Unix
         run: |
           export PATH=$PATH:/home/runner/.mix/elixir/1-17/ # for rebar3 to be found
           mix popcorn.build_runtime --target unix --cmake-opts "SANITIZER=OFF DEBUG_ASSERTIONS=ON DEBUG_GC=ON"
+      - name: Build AtomVM WASM
+        run: |
+          export PATH=$PATH:/home/runner/.mix/elixir/1-17/ # for rebar3 to be found
```
bblaszkow06: To not repeat that, wouldn't it be enough to configure rebar3 via `setup-beam`

README.md
```diff
@@ -75,6 +75,52 @@ Cross-Origin-Embedder-Policy: require-corp
...
+
+Popcorn tests can be run either on WASM via Playwright or natively on UNIX. To run them on WASM, run
+```
+TARGET=wasm mix test
+```
+
+To run tests on UNIX, use
+
+```
+mix popcorn.build_runtime --target unix
```
bblaszkow06: Not sure if this was intended, but for this to work, I had to set `MIX_ENV=test`

PR #183, opened 2025-06-23
Fixed crashes caused by get_object_code implementation

FKubisSWM:
Added a patch to fix the WASM crashes in beam_lib module.
After ParallelChecker got proper module binary from get_object_code function it crashed while chunking the data in beam_lib module on a binary pattern matching.

### DIFF
--- a/patches/estdlib/beam_lib.erl
+++ b/patches/estdlib/beam_lib.erl
@@ -1,0 +1,30 @@
+-module(beam_lib).
+
+
+-record(bb, {pos = 0 :: integer(),
+  bin :: binary(),
+  source :: binary() | string()}).
+
+-compile({popcorn_patch_private, pread/3}).
+
+pread(FD, AtPos, Size) ->
+  #bb{pos = Pos, bin = Binary} = FD,
+  Skip = AtPos-Pos,
+  case Binary of
+%%    Patch reason: 
+%%    The pattern matching binary with < Skip size to:
+%%    "<<_:Skip/binary, B:Size/binary, Bin/binary>>" 
+%%    causes a crash for yet unspecified reasons.
+%%    Patch:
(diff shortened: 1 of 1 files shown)


config/config.exs
```diff
@@ -2,7 +2,9 @@ import Config
 
 config :popcorn,
   add_tracing: false,
-  runtime_source: {:git, "git@github.com:software-mansion-labs/FissionVM.git"}
+  runtime_source:
```
FranekKubis: to be removed - added only for CIs sake

patches/estdlib/beam_lib.erl
```diff
@@ -0,0 +1,30 @@
...
+  source :: binary() | string()}).
+
+-compile({popcorn_patch_private, pread/3}).
+
+pread(FD, AtPos, Size) ->
+  #bb{pos = Pos, bin = Binary} = FD,
+  Skip = AtPos-Pos,
+  case Binary of
+%%    Patch reason: 
+%%    The pattern matching binary with < Skip size to:
```
bblaszkow06, line 15:
```suggestion
%%    The pattern matching binary with size < Skip to:
```

patches/estdlib/beam_lib.erl
```diff
@@ -0,0 +1,30 @@
...
+%%    "<<_:Skip/binary, B:Size/binary, Bin/binary>>" 
+%%    causes a crash for yet unspecified reasons.
```
mat-hek, line 17: Please create an issue

mat-hek (approved):

patches/estdlib/beam_lib.erl
FranekKubis: https://github.com/software-mansion/popcorn/issues/194

PR #184, opened 2025-06-25
Logger

mat-hek:


patches/estdlib/erlang.erl
```diff
@@ -2583,6 +2584,15 @@ binary_to_integer(_Binary, Base) ->
 binary_to_list(_Binary) ->
     erlang:nif_error(undefined).
 
+% Patch reason: NIF nor available in AtomVM
+-spec binary_to_list(Binary, Start, Stop) -> [byte()] when
```
jgonet, line 2588: Typo: "nor" -> "not".

jgonet (approved):
Can you create issues for:
- missing `binary_to_list/3` NIF
- missing `:io.printable_range/0` NIF
- missing `:os.system_time/0` NIF
- missing `messages` key in `process_info`.
- genserver issue:

```elixir
        defmodule GS do
           # FIXME: use GenServer generates 'Unknown external term type: 259'
           # use GenServer

           def init(_opts) do
             {:ok, %{}}


           end

           def handle_info(:exit, _state) do
             raise "foo"
           end
         end
```

PR #185, opened 2025-06-26
Batch JS calls

jgonet:
This uses https://github.com/software-mansion-labs/FissionVM/pull/140, reworking internals of the API to utilize operation on ref batches.

Other changes:
- `args` are extracted to argument instead of being in keyword list. 
- Using Wasm module callbacks, evaluation is moved from AtomVM to `onRunTrackedJs` which allows for debug assertions and changes being done without changing code in the VM.
- `window` argument is removed, now `run_js` uses variable shadowing to replace iframe window with main window.

One unresolved issue is to model values that need destructors (e.g. `setInterval`). This is currently done by generating keys in the JS itself, registering destructors in auxiliary map, and returning `TrackedValue` that is used as container for key and value (to prevent generating new key).

### DIFF
--- a/examples/game_of_life/lib/game_of_life/ui.ex
+++ b/examples/game_of_life/lib/game_of_life/ui.ex
@@ -129,7 +129,6 @@ defmodule GameOfLife.Ui do
     }
     """
     |> Wasm.run_js!(%{ms: ms, receiver: @receiver_name})
-    |> hd()
   end
 
   defp stop_timer(timer_ref) do
@@ -156,7 +155,7 @@ defmodule GameOfLife.Ui do
     Map.new(selectors, fn {key, selector} ->
       node_ref = query_selector(selector)
 
-      {:ok, [listener_ref]} =
+      {:ok, listener_ref} =
         Wasm.register_event_listener(:click,
           target_node: node_ref,
           event_receiver: @receiver_name,
@@ -230,6 +229,5 @@ defmodule GameOfLife.Ui do
     }
     """
     |> Wasm.run_js!(%{selector: selector})
-    |> hd()
   end
 end
--- a/examples/game_of_life/lib/game_of_life/ui.ex
+++ b/examples/game_of_life/lib/game_of_life/ui.ex
@@ -123,14 +123,13 @@ defmodule GameOfLife.Ui do
 
   defp start_timer(ms) do
     # TODO: add unregistration on process exit
-    Wasm.run_js!(
-      """
-      ({ wasm, args }) => {
-        return setInterval(() => wasm.cast(args.receiver, "tick"), args.ms);
-      }
-      """,
-      args: %{ms: ms, receiver: @receiver_name}
-    )
+    """
+    ({ wasm, args }) => {
+      return [setInterval(() => wasm.cast(args.receiver, "tick"), args.ms)];
+    }
+    """
+    |> Wasm.run_js!(%{ms: ms, receiver: @receiver_name})
+    |> hd()
   end
 
   defp stop_timer(timer_ref) do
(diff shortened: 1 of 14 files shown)


examples/hello_popcorn/config/config.exs
```diff
@@ -2,4 +2,5 @@ import Config
 
 config :popcorn,
   start_module: HelloPopcorn,
+  runtime_source: {:path, "../../../AtomVM"},
```
mat-hek: 🤨

examples/hello_popcorn/mix.exs
```diff
@@ -26,4 +29,15 @@ defmodule HelloPopcorn.MixProject do
...
+  defp copy_artefacts(_) do
+    mjs = Application.app_dir(:popcorn, ["atomvm_artifacts", "wasm", "AtomVM.mjs"])
+    wasm = Application.app_dir(:popcorn, ["atomvm_artifacts", "wasm", "AtomVM.wasm"])
+    popcorn_js = Application.app_dir(:popcorn, ["priv", "static-template", "wasm", "popcorn.js"])
+    popcorn_iframe_js = Application.app_dir(:popcorn, ["priv", "static-template", "wasm", "popcorn_iframe.js"])
+    File.cp!(mjs, "static/wasm/AtomVM.mjs")
+    File.cp!(wasm, "static/wasm/AtomVM.wasm")
+    File.cp!(popcorn_js, "static/wasm/popcorn.js")
+    File.cp!(popcorn_iframe_js, "static/wasm/popcorn_iframe.js")
+  end
```
mat-hek: this should be done with `popcorn.cook`

lib/popcorn/api/wasm.ex
```diff
@@ -131,84 +135,113 @@ defmodule Popcorn.Wasm do
...
+  Typically, `window` is used to interact with main browser window context. Popcorn shadows `window` variable for that reason.
+
+  ## Example
+  ```
+  Popcorn.Wasm.run_js(\"\"\"
+  ({ args }) => {
+    const n = args.n;
+    return [n-1, n, n-1];
+  }
+  \"\"\", %{n: 5})
```
mat-hek:
```suggestion
  Popcorn.Wasm.run_js(\"""
  ({ args }) => {
    const n = args.n;
    return [n-1, n, n-1];
  }
  \""", %{n: 5})
```

examples/game_of_life/lib/game_of_life/ui.ex
```diff
@@ -157,7 +156,7 @@ defmodule GameOfLife.Ui do
     Map.new(selectors, fn {key, selector} ->
       node_ref = query_selector(selector)
 
-      {:ok, listener_ref} =
+      {:ok, [listener_ref]} =
```
mat-hek: shouldn't register_event_listener just return a listener_ref?

examples/game_of_life/lib/game_of_life/ui.ex
```diff
@@ -123,14 +123,13 @@ defmodule GameOfLife.Ui do
...
-      """,
-      args: %{ms: ms, receiver: @receiver_name}
-    )
+    """
+    ({ wasm, args }) => {
+      return [setInterval(() => wasm.cast(args.receiver, "tick"), args.ms)];
+    }
+    """
+    |> Wasm.run_js!(%{ms: ms, receiver: @receiver_name})
+    |> hd()
```
mat-hek: TBH, it's a bit annoying always to handle a list. Especially that usually only a single value is returned. Do we have any use cases when more values are returned?

mat-hek (changes requested):

PR #191, opened 2025-07-02
Refactor runtime config

mat-hek:


mix.exs
```diff
@@ -72,7 +72,7 @@ defmodule Popcorn.MixProject do
   end
 
   defp download_artifacts(_args) do
-    Popcorn.Utils.FetchArtifacts.fetch_artifacts()
+    Popcorn.Utils.FetchArtifacts.download_artifacts(force: "--force" in System.argv())
```
FranekKubis, line 78:
```suggestion
    Popcorn.Utils.FetchArtifacts.download_artifacts(force: true in System.argv())
```

mat-hek: Did you check that this works? I can't see how it can work 🤔

FranekKubis: I did not checked if it works, but the `@spec` in Popcorn.Utils.FetchArtifacts line 15 and and if statement in line 22 implies using boolean there.

FranekKubis: Oh I misread that.

lib/popcorn.ex
```diff
@@ -300,4 +300,19 @@ defmodule Popcorn do
 
     gather_app_specs(deps, Map.merge(specs, new_specs))
   end
+
+  defp cp_gzip(paths, out_dir, handle_missing_file \\ fn _path -> :ok end) do
```
jgonet, line 304: Why we need that?

test/popcorn/eval_test.exs
```diff
@@ -672,7 +672,10 @@ defmodule Popcorn.EvalTest do
       |> AtomVM.try_eval(:elixir, run_dir: dir)
 
     assert %{exit_status: 0, output: output} = result
-    assert [debug, info, warning, error] = String.split(output, "\n", trim: true)
+
+    assert [debug, info, warning, error] =
+             String.split(output, "\n", trim: true) |> Enum.reject(&String.starts_with?(&1, "\e"))
```
jgonet: What's the `\e`?

jgonet (approved):

lib/popcorn.ex
mat-hek: Not sure what you mean, it was convenient to have a function that copies and gzips what it copied

test/popcorn/eval_test.exs
mat-hek: For some reason (#105 ?) logger started coloring its output, and there are some lines in form of \e***\n, that indicate start/end of particular color

lib/popcorn.ex
jgonet: I mean why we need to gzip copied things instead of relying on gzip compression of webserver. Other user artifacts  aren't compressed.

mat-hek: Web server doesn't compress anything, you need to provide compressed versions of files. Phoenix has a task for it AFAIK.

bblaszkow06: AFAIK, there's no gzip conversion on the web server. `gzip: true` in `Plug.Static` only instructs to look for `.gz` version.

In Phoenix `phx.digest` handles the compression of assets https://hexdocs.pm/phoenix/Mix.Tasks.Phx.Digest.html

PR #196, opened 2025-07-04
Add a dev server which supports live reloading

Gazler:
A new server has been added, which can be copied into the users directory using `mix popcorn.dev_server`. The dev server differs from the simple server as it sets up a file system watcher using the FileSystem library. This serves two purposes:

1. Whenever a module in the Elixir application changes, `mix popcorn.cook` is called. This uses `System.cmd` as the server runs separately from popcorn, so we can't invoke the mix task directly.
2. Whenver the static directory changes, the websocket sends a reload event, which reloads the web browser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,24 @@ Cross-Origin-Embedder-Policy: require-corp
 
 Otherwise, browsers refuse to run WASM.
 
+### Live reloading
+
+Popcorn also ships with a server that supports like reloading. It can be generated with `mix popcorn.dev_server` which will replace the simple server above.
+
+The index.html requires an extra line to load the required JavaScript.
+
+```html
+<!-- index.html -->
+<html>
+  <script type="text/javascript" src="/dev_server.js"></script>
+  <script type="module">
+      import { Popcorn } from "./wasm/popcorn.js";
+      await Popcorn.init({ onStdout: console.log });
+  </script>
+  <body></body>
+</html>
+ ```
+
 ## Configuration of the runtime
 
 Popcorn runs AtomVM under the hood, and therefore it needs to either download precompied artifacts or compile it from source.
--- a/priv/dev_server.exs
+++ b/priv/dev_server.exs
@@ -1,0 +1,154 @@
+Mix.install([
+  {:bandit, "~> 1.1"},
+  {:websock_adapter, "~> 0.5"},
+  {:file_system, "~> 1.0"}
+])
+
+defmodule Reloader do
+  require Logger
+
+  def init(_args) do
+    Registry.register(Dispatcher, :reload, [])
+    {:ok, []}
+  end
+
+  def handle_in({"ping", [opcode: :text]}, state) do
+    {:reply, :ok, {:text, "pong"}, state}
+  end
+
+  def handle_info(:reload, state) do
+    Logger.debug("mix popcorn.cook finished. Reloading page.")
+    {:push, {:text, "reload"}, state}
+  end
+
+  def handle_info(_, state) do
+    {:ok, state}
+  end
+end
+
+defmodule Router do
+  use Plug.Router
+
+  @static_dir "#{__DIR__}/static"
+
+  plug(Plug.Logger)
+  plug(:add_headers)
+  plug(Plug.Static, from: @static_dir, at: "/")
+  plug(:match)
+  plug(:dispatch)
+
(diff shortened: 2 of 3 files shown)


priv/dev_server.exs
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  def handle_in({"ping", [opcode: :text]}, state) do
+    {:reply, :ok, {:text, "pong"}, state}
+  end
+
+  def handle_info({:file_event, _, {path, [:modified, :closed]}}, state) do
+    if String.contains?(path, "#{__DIR__}/static") do
+      Process.send_after(self(), :reload, 50)
+      {:push, {:text, "reload"}, state}
```
mat-hek: Why do we reload twice? What if compiling takes longer than 50ms? Maybe we could send a message from recompiler to reloader when the compilation is finished, instead of subscribing to `live_reload_file_monitor` in both processes?

priv/dev_server.exs
```diff
@@ -0,0 +1,128 @@
...
+  plug(:add_headers)
+  plug(Plug.Static, from: @static_dir, at: "/")
+  plug(:match)
+  plug(:dispatch)
+
+  get "/" do
+    send_file(conn, 200, "#{@static_dir}/index.html")
+  end
+
+  get "app.js" do
```
mat-hek: How about

```suggestion
  get "dev_server.js" do
```

README.md
```diff
@@ -114,6 +114,24 @@ Cross-Origin-Embedder-Policy: require-corp
...
+### Live reloading
+
+Popcorn also ships with a server that supports like reloading. It can be generated with `mix popcorn.dev_server` which will replace the simple server above.
+
+The index.html requires an extra line to load the required JavaScript.
+
+```html
+<!-- index.html -->
+<html>
+  <script type="text/javascript" src="/app.js"></script>
```
mat-hek:
```suggestion
  <script type="text/javascript" src="/dev_server.js"></script>
```

priv/dev_server.exs
```diff
@@ -0,0 +1,128 @@
...
+  def init(_args) do
+    FileSystem.subscribe(:live_reload_file_monitor)
+    {:ok, []}
+  end
+
+  def handle_in({"ping", [opcode: :text]}, state) do
+    {:reply, :ok, {:text, "pong"}, state}
+  end
+
+  def handle_info({:file_event, _, {path, [:modified, :closed]}}, state) do
```
mat-hek: please name ignored variables

priv/dev_server.exs
```diff
@@ -0,0 +1,128 @@
...
+    {:noreply, state}
+  end
+end
+
+{opts, _argv} = OptionParser.parse!(System.argv(), strict: [port: :integer])
+
+bandit = {Bandit, plug: Router, scheme: :http, port: Keyword.get(opts, :port, 4000)}
+
+file_system =
+  {FileSystem.Worker, name: :live_reload_file_monitor, dirs: Enum.map([""], &Path.absname/1)}
```
mat-hek:
```suggestion
  {FileSystem.Worker, name: :live_reload_file_monitor, dirs: Enum.map(["."], &Path.absname/1)}
```

mat-hek (changes requested):
Hi @Gazler, thanks for the PR! It doesn't work on my Mac, probably because you match on `{:file_event, _, {path, [:modified, :closed]}`, while I see `[:inodemetamod, :modified]` coming when I change a file. I'm testing on `examples/hello_popcorn`

priv/dev_server.exs
Gazler: Yeah, that's probably a better idea, there can be more than one Reloader process, but we can easily do something using the Registry.

Gazler:
I've changed the implementation slightly. It now uses Registry to dispatch reload messages to the Reloader. There's now a lock on the cook job too, it is run asynchronously, and will only cook if it's not already running. This will prevent multiple file changes from making it rebuild constantly. There's a list of valid_events, hopefully one of those will catch your file changes!

mat-hek:
Perfect, it works now! I have two small suggestions, and we can merge:
- IMO it would be good to cook when starting the server
- The generated file shouldn't be named `server.exs`, not to overwrite the simple_server - maybe `dev_server.exs`?

PR #197, opened 2025-07-04
fix invalid github URL

Gazler:
Previously, the GitHub URL linked pointed to swmansion instead of software-mansion. This also sets the source_url used by ExDoc for autolinking to the GitHub repo.


jgonet (approved):
Thanks!

PR #209, opened 2025-07-14
Fix script loading in an iframe

FranekKubis:


priv/static-template/wasm/popcorn.js
```diff
@@ -67,14 +68,15 @@ export class Popcorn {
         "Don't construct the Popcorn object directly, use Popcorn.init() instead",
       );
     }
-    const { bundlePath, onStderr, onStdout, heartbeatTimeoutMs, debug } =
+    const { bundlePath, onStderr, onStdout, heartbeatTimeoutMs, debug, wasmDir } =
```
jgonet, line 72: Did you experiment with other solutions? We could potentially assume that both scripts are in the same locations and base the path of it.

FranekKubis: I do not really understand - right now we are assuming every script (popcorn.js, AtomVM.mjs etc) is on the same location which was hard coded and now it is possible to change this location.
Or do you mean to generate the scripts into another location?

jgonet: No, I meant that you can retrieve where the script is located 
https://github.com/software-mansion/popcorn/blob/58bfcfc0de602d46a266887372b6008146d0f074/fission_lib/priv/static-template/wasm/fission.js#L85-L90
This particular solution was not working but I didn't have time to investigate more.

If we decide to add `wasmDir` we need to at least document it in some way.

FranekKubis: I would stay with the wasmDir approach - I'll add some documentation for that.

jgonet (approved):

PR #218, opened 2025-07-16
Fixing the "noproc" error

FranekKubis:


priv/static-template/wasm/popcorn_iframe.js
```diff
@@ -149,6 +149,11 @@ function onVmInit(initProcess) {
         const result = await Module.call(process, args);
         send(MESSAGES.CALL, { requestId, data: Module.deserialize(result) });
       } catch (error) {
+        if(error == "noproc") {
+          send(MESSAGES.RELOAD, null);
+          console.error("Runtime VM crushed, popcorn iframe reloaded.");
```
mat-hek: to 'crush' means to 'compress or squeeze forcefully so as to break, damage, or distort in shape.' :P

```suggestion
          console.error("Runtime VM crashed, popcorn iframe reloaded.");
```

PR #223, opened 2025-07-18
IEx Wasm: fix shell type

mat-hek:


jgonet:
Fixed in https://github.com/software-mansion/popcorn/pull/224

PR #232, opened 2025-07-25
Fix landing page styling

mmietla:


misc/landing-page/src/styles/global.css
```diff
@@ -5,6 +5,8 @@
   --color-white: #fff;
   --color-light-20: #fffdf5;
   --color-light-30: #fffaf7;
+  --color-grey-20: #efecea;
+  --color-orange-20: #ffe6cc;
```
jgonet, line 9: We can remove `--color-orange-weird`, we know its name now.

misc/landing-page/src/components/Steps.astro
```diff
@@ -37,3 +41,89 @@ import Section from "./Section.astro";
     >
   </Step>
 </Section>
+
+<script>
+  function setupStepAnimation() {
```
jgonet: I think we should move it to new PR, it behaves differently than reference (entire content replacing instead of individual elements, numbers scrolling separately) and scrolling from header works weirdly.

misc/landing-page/src/components/Info.astro
```diff
@@ -7,21 +7,23 @@ import Section from "./Section.astro";
...
+        >Apache 2.0</a
       >. You’re free to use it in personal and commercial projects.
     </p>
     <p>
       Need help with something bigger? We’re <a
         target="_blank"
         rel="noopener noreferrer"
-        href="https://swmansion.com"
-        class="underline">Software Mansion</a
+        href="https://swmansion.com">Software Mansion</a
```
jgonet: Should be underlined, no?

misc/landing-page/src/components/LiveDemo.astro
```diff
@@ -8,7 +8,12 @@ const exampleButton =
   "flex items-center gap-2 text-orange-100 cursor-pointer font-semibold";
 ---
 
-<Section id="live-demo" background="light" class="hidden lg:flex">
+<Section
+  id="live-demo"
+  background="light"
+  class="hidden lg:flex"
+  outerClass="overflow-x-hidden"
```
jgonet: Can we achieve this somehow else?

misc/landing-page/src/components/Marquee.astro
```diff
@@ -74,9 +78,9 @@ const marqueeElementClass = "marquee-element flex flex-shrink-0";
 
       lastTimestamp = timestamp;
 
-      // Apply combined transformation (scroll effect + continuous animation)
+      // Apply transformation
```
jgonet: Useless comment.

misc/landing-page/src/components/Info.astro
mmietla: For the design, we changed the underline to orange text color

misc/landing-page/src/components/LiveDemo.astro
mmietla: I've been thinking about it, and I can't figure out a simpler approach. We could either move the marquee element higher, to the same level as the section, or overwrite the default section margin classes and apply them to the elements except the marquee.

mmietla: I found the CSS overflow property 'clip'. When applied to the main layout tag, it works fine.

PR #248, opened 2025-07-31
GameOfLife: stop and restart buttons

bblaszkow06:


examples/game_of_life/lib/game_of_life/ui.ex
```diff
@@ -113,7 +117,11 @@ defmodule GameOfLife.Ui do
     """
     <div class="controls">
       <button id="start">Start simulation</button>
-      <button id="glider">Use glider preset</button>
+      <button id="stop" hidden>Stop simulation</button>
+      <button id="reset">Reset</button>
```
jgonet, line 121: Can't we reuse "start" for stopping?

bblaszkow06: We'd have to change the text then or use something like `Start/stop simulation`. IMO this is cleaner as we have separate events for start and stop, plus the HTML content is in one place.

jgonet: That's fine then. UX-wise I think one button is better (e.g. what happens if you hit "start" when simulation is running) but we can come back to it in the future.

mmietla: If possible, we can also disabled buttons based on current state of the game of life. IMO this will result better UX

Issue #237, opened 2025-07-28
Add page templating for lang tour


jgonet:
Lang tour pages should be able to be built from some templates/components to easily adapt from markdown guide pages to lang tour version.

Pages should be grouped by category/main concept (e.g. "Processes" that contain send/receive, genservers, queue, tasks, etc).

PR #256, opened 2025-08-12
Lang tour templating and navigation

mmietla:
implements #237 #238 

https://github.com/user-attachments/assets/1c2017ae-2483-4b5d-9292-c8bffb1ee74e

### DIFF
--- a/misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
+++ b/misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
@@ -5,20 +5,23 @@ import ChevronDown from "../../assets/chevron-down.svg?react";
 import ChevronRight from "../../assets/chevron-right.svg?react";
 import { useState } from "react";
 
-export function NavigationItem({
-  item,
-  onClick
-}: {
+type NavigationItemProps = {
   item: NavigationTreeItem;
   onClick?: () => void;
-}) {
+};
+
+export function NavigationItem({ item, onClick }: NavigationItemProps) {
   const { hash, pathname } = useLocation();
 
   const [isOpen, setIsOpen] = useState(
-    `${pathname}${hash}`.includes(item.path || "")
+    `${pathname}${hash}`.includes(item.path ?? "")
   );
 
-  const hasChildren = item.children && Object.keys(item.children).length > 0;
+  const hasChildren = item.children !== undefined;
+
+  const isSelected = (item: NavigationTreeItem) => {
+    return item.path && item.path.includes(hash);
+  };
 
   return (
     <li className="my-1">
@@ -28,7 +31,7 @@ export function NavigationItem({
             ? "text-brown-header hover:bg-orange-20 mt-2 cursor-pointer font-semibold"
             : "text-brown-80"
         }`}
-        onClick={() => hasChildren && setIsOpen(!isOpen)}
+        onClick={() => hasChildren && setIsOpen((prev) => !prev)}
       >
         {hasChildren && (
           <span className="mr-2 text-xs text-orange-100">
@@ -44,7 +47,7 @@ export function NavigationItem({
             to={item.path}
             className={({ isActive }) =>
               `flex w-full items-center rounded-md px-3 py-0.5 text-sm transition-colors duration-200 ${
-                isActive && item.path && item.path.includes(hash)
+                isActive && isSelected(item)
                   ? "bg-color-orange-20 text-color-orange-100 font-semibold"
                   : "hover:bg-orange-20 text-color-brown-80 hover:text-color-brown-100"
               } ${hasChildren ? "pl-0" : "pl-5"}`
@@ -62,7 +65,7 @@ export function NavigationItem({
         <ul
           className={`pl-5 transition-all duration-300 ${isOpen ? "block opacity-100" : "hidden opacity-0"}`}
         >
-          {Object.entries(item.children || {}).map(([key, childItem]) => (
+          {Object.entries(item.children ?? {}).map(([key, childItem]) => (
             <NavigationItem key={`${key}`} item={childItem} onClick={onClick} />
           ))}
         </ul>


misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,72 @@
+import { NavLink, useLocation } from "react-router";
+import type { NavigationTreeItem } from "../../utils/content/navigation";
+
+import ChevronDown from "../../assets/chevron-down.svg?react";
+import ChevronRight from "../../assets/chevron-right.svg?react";
```
jgonet, line 5: We could also use some package for icons as well.

misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { useState } from "react";
+
+export function NavigationItem({
+  item,
+  onClick
+}: {
+  item: NavigationTreeItem;
+  onClick?: () => void;
+}) {
```
jgonet: Hint: if you use `type Props = ` above component, it makes it a single line instead of four lines.

misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { hash, pathname } = useLocation();
+
+  const [isOpen, setIsOpen] = useState(
+    `${pathname}${hash}`.includes(item.path || "")
```
jgonet: Maybe?
```suggestion
    `${pathname}${hash}`.includes(item.path ?? "")
```

misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+  const hasChildren = item.children && Object.keys(item.children).length > 0;
```
jgonet: Not going to force it on you but I personally avoid doing implicit coertion to bools and use full `!== undefined` (this also avoid ambiguity between null and undefined).

misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <li className="my-1">
+      <div
+        className={`flex items-center rounded-md px-3 py-0.5 text-sm ${
+          hasChildren
+            ? "text-brown-header hover:bg-orange-20 mt-2 cursor-pointer font-semibold"
+            : "text-brown-80"
+        }`}
+        onClick={() => hasChildren && setIsOpen(!isOpen)}
```
jgonet: I'm not sure about rendering pipeline but I vaguely remember this is a good candidate to use `setState(prev => ...)` instead of using previous state from `const [state, setState] = useState(...)`

misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+              <ChevronRight className="w-5" />
+            )}
+          </span>
+        )}
+        {item.path ? (
+          <NavLink
+            to={item.path}
+            className={({ isActive }) =>
+              `flex w-full items-center rounded-md px-3 py-0.5 text-sm transition-colors duration-200 ${
+                isActive && item.path && item.path.includes(hash)
```
jgonet: IMO this is too much logic in the rendering part. Maybe `isSelected(item)`?

misc/language-tour-guide/src/components/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+        ) : (
+          <span className="capitalize">{item.title}</span>
+        )}
+      </div>
+
+      {hasChildren && (
+        <ul
+          className={`pl-5 transition-all duration-300 ${isOpen ? "block opacity-100" : "hidden opacity-0"}`}
+        >
+          {Object.entries(item.children || {}).map(([key, childItem]) => (
```
jgonet:
```suggestion
          {Object.entries(item.children ?? {}).map(([key, childItem]) => (
```

misc/language-tour-guide/src/components/navigation/ScrollToHash.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const { hash } = useLocation();
+
+  useEffect(() => {
+    if (hash) {
+      const id = hash.slice(1);
+      const el = document.getElementById(id);
+      if (el) {
+        setTimeout(() => {
+          el.scrollIntoView();
+        }, 100);
```
jgonet: Don't we need cleanup function for it? Why timeout is needed here?

misc/language-tour-guide/src/components/navigation/ScrollToHash.tsx
```diff
@@ -0,0 +1,27 @@
...
+      }
+    } else {
+      const descriptionSection = document.getElementById("description");
+      if (!descriptionSection) {
+        return;
+      }
```
jgonet: In block above you do the logic inside if and in this block you reverse it. Please choose the same approach in both cases.

misc/language-tour-guide/src/components/navigation/ScrollToHash.tsx
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from "react";
+import { useLocation } from "react-router";
+
+export function ScrollToHash() {
```
jgonet: I don't really get why this is a component.

misc/language-tour-guide/src/AppRoutes.tsx
```diff
@@ -1,12 +1,16 @@
 import { Route, Routes } from "react-router";
-import { Home } from "./pages/Home";
 import { Layout } from "./components/Layout";
+import { ExampleLayout } from "./components/ExampleLayout";
+import { createRouteComponents } from "./utils/content/content-routes";
+import { useMemo } from "react";
 
 export function AppRoutes() {
+  const contentRoutes = useMemo(() => createRouteComponents(), []);
```
jgonet: Nit: Or `useMemo(createRouteComponents, []);`

misc/language-tour-guide/src/utils/content/index.ts
```diff
@@ -0,0 +1,18 @@
+import type { ComponentType } from "react";
+
+const regex = /^\/src\/content(\/[^.]+)(?:\.[^.]+)$/;
```
jgonet: I feel like this regex could be simpler, what we want to match?

misc/language-tour-guide/src/pages/Home.tsx
```diff
@@ -1,51 +1,15 @@
...
-          tempora quibusdam! Consectetur molestiae architecto nemo voluptate,
-          rem tempora, mollitia numquam, accusantium reprehenderit veniam amet
-          facere? This is an example layout for the Language Tour Guide. It
-          includes a header and a description section. Lorem ipsum dolor sit
-          amet consectetur adipisicing elit. Cumque, eveniet autem odit nam
-          nostrum tempora quibusdam! Consectetur molestiae architecto nemo
-          voluptate, rem tempora, mollitia numquam, accusantium reprehenderit
-          veniam amet facere?
-        </p>
+        {/* <Agent /> */}
```
jgonet: Leftovers?

misc/language-tour-guide/src/components/Loader.tsx
```diff
@@ -0,0 +1,13 @@
+interface LoaderProps {
+  message?: string;
+}
```
jgonet: We should use one solution, right now we have some props inline, this has interface and I think I saw `type` used as well.

misc/language-tour-guide/src/utils/content/navigation.ts
```diff
@@ -0,0 +1,96 @@
+import { getRoutePath, mdxModules } from ".";
+
+export type NavigationTreeItem = {
+  title?: string;
+  path?: string;
+  children?: NavigationTree;
+  order?: number;
+  type: "section" | "subsection";
```
jgonet: Why title and path are optional?

misc/language-tour-guide/src/utils/content/navigation.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+export type NavigationTree = Record<string, NavigationTreeItem>;
+
+export async function createNavigation(): Promise<NavigationTree> {
```
jgonet, line 13: This could be written better IMO.

misc/language-tour-guide/src/pages/Home.tsx
mmietla: yes

misc/language-tour-guide/src/components/navigation/ScrollToHash.tsx
mmietla: I am inspired by this [article ](https://dev.to/mindactuate/scroll-to-anchor-element-with-react-router-v6-38op), but you are right, it can be a custom hook.

misc/language-tour-guide/src/components/navigation/ScrollToHash.tsx
mmietla: It is leftovers from the { behavior: 'smooth' } option in scrollIntoView to make non-instant scrolling to the element. I will remove it.

misc/language-tour-guide/src/utils/content/navigation.ts
mmietla: Created issue for refactor #257

PR #260, opened 2025-08-14
Dynamically include beams of OTP applications

bblaszkow06:


lib/popcorn/build.ex
```diff
@@ -6,9 +6,27 @@ defmodule Popcorn.Build do
...
+    :kernel,
+    :public_key,
+    :stdlib,
+    :ssl,
+    # Elixir
+    :elixir,
+    :eex,
+    :logger,
+    :iex
+  ]
```
jgonet: You can add a comment how this list was created in case we don't support all apps erlang ships or new one is added.

lib/popcorn/build.ex
```diff
@@ -18,46 +36,112 @@ defmodule Popcorn.Build do
...
+      |> Map.new()
+
+    patches_srcs = Path.wildcard("patches/**/*.{ex,erl,yrl,S}") |> MapSet.new()
 
     cache =
       with false <- opts.force,
            {:ok, cache} <- File.read(@cache_path),
            cache = :erlang.binary_to_term(cache),
+           cached_files =
+             cache |> Map.values() |> Enum.reduce(&Map.merge/2) |> MapSet.new(&elem(&1, 0)),
```
mat-hek: A bit too magical for me 🧙

lib/popcorn/build.ex
mat-hek: Ideally, we wouldn't maintain this list and build all apps available, filter out unsupported ones if needed

lib/popcorn/build.ex
```diff
@@ -3,61 +3,172 @@ defmodule Popcorn.Build do
...
+    elixir_apps =
+      Path.expand("..", Application.app_dir(:elixir))
+      |> File.ls!()
+
+    otp_apps ++ elixir_apps
+  end
 
-    :ok
+  @doc """
+  Returns a list of applicatons that were included during compilation
```
jgonet: "applicatons" -> "applications"

lib/popcorn/build.ex
bblaszkow06: Should be better now

lib/popcorn.ex
```diff
@@ -96,14 +91,36 @@ defmodule Popcorn do
...
+
+      missing ->
+        raise CookingError, """
+        Popcorn was built without the following apps: #{inspect(missing)}
+        Add them to :extra_apps in your config.exs:
+
+          config :popcorn, extra_apps: #{inspect(missing)}
+
+        and recompile the project
+        """
```
mat-hek, line 115: Ideally we would build these apps on demand, but probably not worth the effort

Issue #239, opened 2025-07-28
Add evaluation logic in the Elixir side


jgonet:
Evaluation button should be working, with result arena that shows result of evaluation and timings. We should consider presenting result history in some way. Test this extensively with things like exceptions, modules, etc. Missing features should present adequate error instead of crashing the VM.

PR #268, opened 2025-08-19
Lang tour evaluation

mmietla:
Implements #239 

image


misc/language-tour-guide/src/components/code-editor/CodeEditor.styles.css
```diff
@@ -5,7 +5,9 @@
 
 @media (width >= 64rem) {
   .cm-editor {
-    height: calc(100vh - 0.3 * (100vh - 68px) - 68px);
+    height: calc(
+      100vh - 0.4 * (100vh - 68px - 48px - 17px) - 68px - 48px - 17px
```
jgonet: What are those pixels?

misc/language-tour-guide/src/components/code-editor/CodeEditor.tsx
```diff
@@ -1,38 +1,27 @@
...
-  const onChange = useCallback((val: string) => {
-    console.log("val:", val);
-    setValue(val);
-  }, []);
+  const onChange = useCallback(
+    (val: string) => {
+      setCode(val);
+    },
+    [setCode]
+  );
```
jgonet: This doesn't do anything.

misc/language-tour-guide/src/components/execution/History.tsx
```diff
@@ -0,0 +1,83 @@
...
+    setShowHistory((prev) => !prev);
+  };
+
+  const formatTime = (date: Date) => {
+    return date.toLocaleTimeString([], {
+      hour: "2-digit",
+      minute: "2-digit",
+      second: "2-digit"
+    });
+  };
```
jgonet: I feel like this can be standalone function. We could have a `formatTimestamp` and `formatDuration` too (right now duration is formatted inline and it feels inconsistent with formatting time).

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return () => {
+      document.removeEventListener("keydown", onKeyDown);
+    };
+  }, [onKeyDown]);
+
+  return (
+    <section className="bg-light-30 border-grey-20 scrollbar min-h-60 overflow-y-scroll rounded-md border pb-4">
+      <div className="flex w-full flex-wrap justify-end gap-3 border-b border-orange-100 py-3 pr-6">
+        <Button title="Reset Code" type="secondary" onClick={handleResetCode} />
+        {/* <Button title="Format Code" type="secondary" /> */}
```
jgonet: A comment.

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+        // Add to history
+        addHistoryEntry({
+          timestamp: new Date(),
+          code,
+          result: data,
+          durationMs
+        });
+      } catch (error) {
+        setPending(false);
```
jgonet: Tip: I try to minimize code in the try-catch, mostly so you can see what can throw and what is merely dependent on that value. Also, sometimes making a wrapper for things that needs to happen always is nice:

```ts
function usePending() {
    const [pending, setPending] = useState<boolean>(false);
    const withPending = useCallback(async (fn: () => Promise<void>) => {
        setPending(true);
        await fn();
        setPending(false);
    }, [setPending]);
    
    return [pending, withPending];
}
```

so in this case, I'd probably write it as:
```ts
function callPopcorn(code: string) {
    let result: TYPE | null = null;
    try {
        result = await call(["eval_elixir", code], { timeoutMs: 10_000 });
    } catch (error) {
        console.error("failed to initialize elixir tour:", error);
    }
    if (result === null) { return; }
    
    const { data, durationMs } = result;
    setResultData(data);
    setDurationMs(durationMs);
    addHistoryEntry({
        timestamp: new Date(),
        code,
        result: data,
        durationMs
    });
}
```

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -0,0 +1,111 @@
...
+        setPending(true);
+        const { data, durationMs } = await call(["eval_elixir", code], {
+          timeoutMs: 10_000
+        });
+
+        setPending(false);
+        setResultData(data);
+        setDurationMs(durationMs);
+
+        // Add to history
```
jgonet: This is unnecessary, `addHistoryEntry` doesn't need explanation that it adds to history.

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -0,0 +1,111 @@
...
+      if ((e.metaKey || e.ctrlKey) && e.key === "Enter") {
+        handleRunCode();
+      }
+    },
+    [handleRunCode]
+  );
+
+  const handleResetCode = useCallback(() => {
+    resetToDefault();
+  }, [resetToDefault]);
```
jgonet: Do you need to cache a function that is coming from the store? It should be already not changing.

misc/language-tour-guide/src/components/store/codeEditor.ts
```diff
@@ -0,0 +1,27 @@
...
+  resetToDefault: () => void;
+};
+
+export const useCodeEditorStore = create<CodeEditorStore>((set) => ({
+  defaultCode: "",
+  code: "",
+  stdoutResult: [],
+  setCode: (code: string) => set({ code }),
+  setDefaultCode: (defaultCode: string) =>
+    set({ defaultCode, code: defaultCode }),
```
jgonet: Why it changes default code and current code?

misc/language-tour-guide/src/components/markdown/MarkdownComponents.tsx
```diff
@@ -63,14 +64,43 @@ export const BlockQuote = ({
...
+  }, [rawCode, setCode]);
+
+  return (
+    <pre
+      className="text-brown-90 bg-grey-10 relative my-4 overflow-x-auto rounded p-4 font-mono text-sm"
+      {...props}
+    >
+      {children}
+      <button
+        onClick={handleUseInEditor}
```
jgonet: We shouldn't have a button that does nothing sometimes.

jgonet (changes requested):

misc/language-tour-guide/src/components/code-editor/CodeEditor.styles.css
mmietla: There are heights for the header, navigation, and gap. I need those heights to display the CodeMirror with overflow scrolling correctly

misc/language-tour-guide/src/components/store/codeEditor.ts
mmietla: I wanted to set code, because on initial render when defaultCode will be set also code will be display on page with this initial default code.

jgonet: I think we can just call `setCode` along then.

misc/language-tour-guide/src/components/code-editor/CodeEditor.styles.css
jgonet: Hm, can we wrap the editor so it fills its parent to full and just control containers position? Current approach feels extremely brittle.

mmietla: I was trying to make this more flexible, but I can't figure out how because the CodeMirror element is behaving strangely. I’ll think about it one more time

PR #281, opened 2025-08-30
Add issue templates

jgonet:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,15 @@
...
+assignees: ''
+
+---
+
+## Description
+
+## Info
+<!-- Use commit hash or released version -->
+**Affected version:**
+**Is code compiled or interpreted:**
```
bblaszkow06, line 15: How about something more elaborate:
```suggestion
## Expected Behaviour
<!--- What should happen? -->

## Current Behaviour
<!--- What happens instead of the expected behaviour? -->

## Steps to Reproduce
<!--- Provide an unambiguous set of steps to help us reproduce this bug. -->
<!--- Include code snippet or a link to repo/gist if possible -->
1.
2.
3.

## Context (Environment)

- **Popcorn version**: <!-- Use commit hash or released version -->
- **Runtime source/version**:
- **Browser/OS:**
- **Is the code compiled or interpreted:**
<!-- Add any other info that might be relevant -->
```

jgonet: I don't like elaborate issue templates because people don't follow it. I can change Popcorn version to "Popcorn and AtomVM version" but I feel like this is asking too much for people to sieve through two repos (and they need to know the difference between those two).

We can add this info internally but publicly I think current version is super good enough for our purposes.

jgonet: A good example of templates demanding too much is issue in Go written by Rob Pike: `https://github.com/golang/go/issues/45624` (search in page for "Would you consider yourself a novice, intermediate, or experienced Go programmer?")

PR #284, opened 2025-09-01
Move docs to ex_doc

bblaszkow06:
Hosted version can be seen at: https://dev.bblaszkow.pl/popcorn_doc/

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -151,7 +151,7 @@ defmodule Popcorn.MixProject do
       {:dialyxir, ">= 0.0.0", only: [:dev, :test], runtime: false},
       {:credo, ">= 0.0.0", only: [:dev, :test], runtime: false},
       # Docs
-      {:ex_doc, "~> 0.34", only: :dev, runtime: false, warn_if_outdated: true},
+      {:ex_doc, "~> 0.34", only: [:dev, :test], runtime: false, warn_if_outdated: true},
       {:makeup_html, ">= 0.0.0", only: :dev, runtime: false, warn_if_outdated: true},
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -69,13 +70,72 @@ defmodule Popcorn.MixProject do
...
+          });
+        }
+      });
+    </script>
+    """
+  end
+
+  defp generate_js_docs(_) do
+    Mix.shell().cmd(
+      "npx documentation build priv/static-template/wasm/popcorn.js -f html -o doc/js-api"
```
bblaszkow06, line 135: This is what Phoenix uses. I've tested `js-doc` as well (results are here https://dev.bblaszkow.pl/popcorn_doc/js-doc/index.html), but I've decided it's way worse, at least without any customisation.

If someone is willing to handle the theming/customisation or knows a better alternative, I'm open to changing that

jgonet: I found [typedoc](https://typedoc.org). Looks a bit like apple's docs and integrates with typescript. Example: https://typedoc.org/example/classes/Customer.html

bblaszkow06: Looks nice, the problem is we don't have typescript :stuck_out_tongue: I mean I've tested it and it complains about lack of tsconfig:
```
[warning] The entry point ./priv/static-template/wasm/popcorn.js is not referenced by the 'files' or 'include' option in your tsconfig
[error] Unable to find any entry points. See previous warnings
```

If we want to migrate to that, I'd leave it as a separate issue to migrate to TS and add typedoc

Issue #278, opened 2025-08-26
Add code saving in editor


mmietla:
Add code saving in the editor to prevent from progress loss when the user switches to a different one

PR #292, opened 2025-09-04
Lang tour history and code saving

mmietla:
implements: #291, #278 

image

### DIFF
--- a/misc/language-tour-guide/src/components/execution/History.tsx
+++ b/misc/language-tour-guide/src/components/execution/History.tsx
@@ -3,6 +3,7 @@ import { useMemo, useState } from "react";
 import ChevronDownIcon from "../../assets/chevron-down.svg?react";
 import ChevronUpIcon from "../../assets/chevron-right.svg?react";
 import { useExecutionHistoryStore } from "../../store/executionHistory";
+import StdoutResults from "./StdoutResults";
 
 function formatTimeStamp(date: Date): string {
   return date.toLocaleTimeString([], {
@@ -34,9 +35,9 @@ export function History() {
   }, [history]);
 
   return (
-    <div className="mt-6">
+    <div className="border-t border-orange-100">
       <div
-        className="flex cursor-pointer items-center justify-between border-t border-orange-100 pt-3"
+        className="mt-3 flex cursor-pointer items-center justify-between px-6"
         onClick={toggleHistory}
       >
         <h3 className="text-brown-90 text-sm font-medium">Execution History</h3>
@@ -50,7 +51,7 @@ export function History() {
       </div>
 
       {showHistory && (
-        <div className="mt-2 space-y-2">
+        <div className="mt-2 space-y-2 px-6">
           {history.length === 0 ? (
             <p className="text-grey-60 text-xs italic">
               No execution history yet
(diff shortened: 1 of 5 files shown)


misc/language-tour-guide/src/components/execution/History.tsx
```diff
@@ -73,8 +79,30 @@ export function History() {
...
+                          <span
+                            key={`stdout-${index}-${line}`}
+                            className="text-brown-90/70 text-xs font-medium"
+                          >
+                            {line}
+                          </span>
+                        ))}
+                      </div>
+                    </div>
+                  )}
```
bblaszkow06: Maybe it would be worth it to extract as a separate component with stdoutResult passed as property

bblaszkow06 (approved):

misc/language-tour-guide/src/components/code-editor/CodeEditor.tsx
```diff
@@ -4,18 +4,25 @@ import { solarizedLight } from "@uiw/codemirror-theme-solarized";
...
 import { useCodeEditorStore } from "../../store/codeEditor";
+import { useLocation } from "react-router";
 
 export function CodeEditor() {
+  const { pathname } = useLocation();
   const code = useCodeEditorStore((state) => state.code);
   const setCode = useCodeEditorStore((state) => state.setCode);
 
+  const handleChangeCode = (value: string) => {
+    localStorage.setItem(`code-${pathname}`, value);
```
jgonet, line 15: I feel like we should hash this to have shorter keys. Just a thought.

We can also consider to store it in Zustand.

misc/language-tour-guide/src/components/execution/History.tsx
```diff
@@ -73,8 +80,14 @@ export function History() {
                     </span>
                   </div>
                   <div className="rounded bg-white p-2 text-sm whitespace-pre-wrap">
+                    <div className="mb-1">
+                      <span className="text-grey-70 text-xs font-medium">
+                        Result:
+                      </span>
+                    </div>
                     {entry.result || "No result"}
```
jgonet, line 88: `??`

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -1,28 +1,47 @@
...
-  const stdoutResult = useCodeEditorStore((state) => state.stdoutResult);
-  const resetStdoutResult = useCodeEditorStore(
-    (state) => state.resetStdoutResult
-  );
   const addHistoryEntry = useExecutionHistoryStore(
     (state) => state.addHistoryEntry
   );
 
-  const { call } = usePopcorn();
+  const stdoutRef = useRef<string[]>(stdoutResult);
```
jgonet, line 40: Why we need this ref?

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -70,6 +91,11 @@ export function Results() {
     [handleRunCode]
   );
 
+  useEffect(() => {
+    setResultData(null);
+    setDurationMs(null);
+  }, [pathname]);
```
jgonet, line 97: This should be in navigation hook IMO, something like `onNavigationChange` option or something.

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -79,33 +105,41 @@ export function Results() {
...
+            <StdoutResults stdout={stdoutResult} />
+            {/* {stdoutResult.map((line, index) => (
               <span
                 key={`stdout-${index}-${line}`}
                 className="text-brown-90/70 text-xs font-medium"
               >
                 {line}
               </span>
-            ))}
+            ))} */}
```
jgonet, line 138: Do we plan on restoring this in the future? If not, remove.

misc/language-tour-guide/src/components/execution/StdoutResults.tsx
```diff
@@ -0,0 +1,25 @@
+type StdoutResultsProps = {
+  stdout?: string[];
+};
+
+export default function StdoutResults({ stdout }: StdoutResultsProps) {
+  if (!stdout || stdout.length === 0) {
+    return null;
+  }
```
jgonet, line 8: Not a fan, IMO it should be conditionally rendered in parent component.

misc/language-tour-guide/src/components/execution/Results.tsx
mmietla: Without this ref, history saves the previous stdout, because stdout changes when the callPopcorn function runs and the component does not re-render.

misc/language-tour-guide/src/components/execution/Results.tsx
mmietla: To create this hook, I need to either pass the setState function for resultData and durationMs, or create these states inside the hook and return them. Is there any better approach?

PR #296, opened 2025-09-05
Lang tour refactor nav

mmietla:
image

image

### DIFF
--- a/misc/language-tour-guide/src/store/navigation.ts
+++ b/misc/language-tour-guide/src/store/navigation.ts
@@ -1,6 +1,6 @@
-import { create } from 'zustand';
-import { createNavigation } from '../utils/content/navigation-builder';
-import type { NavigationTree } from '../utils/content/types';
+import { create } from "zustand";
+import { createNavigation } from "../utils/content/navigation-builder";
+import type { NavigationTree } from "../utils/content/types";
 
 interface NavigationStore {
   navigation: NavigationTree | null;
@@ -13,13 +13,13 @@ export const useNavigationStore = create<NavigationStore>((set, get) => ({
   isLoading: false,
   loadNavigation: async () => {
     if (get().navigation) return;
-    
+
     set({ isLoading: true });
     try {
       const navTree = await createNavigation();
       set({ navigation: navTree, isLoading: false });
     } catch (err) {
-      console.error('Failed to load navigation:', err);
+      console.error("Failed to load navigation:", err);
       set({ isLoading: false });
     }
   }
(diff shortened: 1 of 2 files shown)


misc/language-tour-guide/src/store/navigation.ts
```diff
@@ -0,0 +1,26 @@
...
+    set({ isLoading: true });
+    try {
+      const navTree = await createNavigation();
+      set({ navigation: navTree, isLoading: false });
+    } catch (err) {
+      console.error('Failed to load navigation:', err);
+      set({ isLoading: false });
+    }
+  }
+}));
```
bblaszkow06: nit: There's a newline character missing at the end of this file. Ensure your editor is configured to insert it automatically

misc/language-tour-guide/src/utils/content/types.ts
```diff
@@ -28,7 +27,7 @@ export type UnresolvedEntry = [string, () => Promise<MdxWithProperties>];
 
 export type DirTree = Map<Dir, DirEntry>;
 
-type NavigationTreeItemType = "link" | "section" | "subsection";
+type NavigationTreeItemType = "link" | "section";
```
bblaszkow06, line 30: I see you have removed the code related to subsections completely. Will any of this be helpful if we want to organise the content into groups?

bblaszkow06 (approved):
Looks good and works, nice job!

misc/language-tour-guide/src/utils/content/types.ts
mmietla: We can still group the content by putting it into a directory

misc/language-tour-guide/src/utils/content/navigation-builder.ts
```diff
@@ -84,6 +69,16 @@ export function getNodeNavigationSiblings(tree: NavigationTree, path: string) {
 
   traverse(tree);
 
+  const parentPath = path.split("/").slice(0, -1).join("/");
+
+  const children = flatNavigation.filter((item) => {
+    const itemParentPath = item.path.split("/").slice(0, -1).join("/");
```
jgonet, line 75: This doesn't handle query params and is pretty opaque in what it does. Can you add a function that parses it to URL, drops query params and does thing above?

We can also preprocess items if it isn't that much work so we don't do this filtering constantly.

misc/language-tour-guide/src/utils/hooks/useSiblingsNode.ts
```diff
@@ -0,0 +1,23 @@
...
+  const navigationProperty = useMemo(() => {
+    if (!navigation) {
+      return {
+        siblingsNode: null,
+        childrenCount: 0,
+        currentIndex: 0
+      };
+    }
+    const result = getNodeNavigationSiblings(navigation, pathname);
+    return result;
```
jgonet, line 19: Nit: just return.

PR #297, opened 2025-09-09
Lang tour content structure

mmietla:
- Created a structure for the language tour guide content.
- Implemented directory ordering in navigation using a numeric prefix.


misc/language-tour-guide/src/utils/content/tree-builder.ts
```diff
@@ -10,28 +10,42 @@ export async function getRawMdxTree() {
...
   }
 
   return tree;
 }
+
+function getFormatDirWithOrder(name: string): [string, number] {
+  const orderMatch = name.match(/^(\d+)-/);
+  const order = orderMatch ? parseInt(orderMatch[1], 10) : Infinity;
+
+  const formatName = name.replace(/^\d+-/, "");
```
bblaszkow06: If you need a single match of regex, you may prefer [exec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) 
If you want to use regexes for that, you can reduce that to one:
```suggestion
  const result = name.match(/^(\d+)-(.*)/);
  const order = result ? parseInt(result[1], 10) : Infinity;
  const formatName = result ? result[2] : name;
```

misc/language-tour-guide/src/content/3-constructs/if-unless.mdx
```diff
@@ -0,0 +1,64 @@
+export const frontmatter = {order: 1}
+
+# If and Unless - Simple Conditionals
```
bblaszkow06: `unless` is deprecated, I'd remove it right away

Issue #128, opened 2025-05-22
Events API


jgonet:
We'd like to allow Elixir to send events to JS side. This would also fix problems with process registration.

Sketch of the API:

```js
// JS
const popcorn = new Popcorn({ ... });
// or onMessage(name, handler) to not listen on all messages
popcorn.onMessage((eventName, payload) => {
  switch(eventName) {
    case "processing_finished":
      handleProcessingFinished(payload.id);
      break;
    case "popcorn_register_receiver":
      if(popcorn.defaultReceiver === null) {
        popcorn.setDefaultReceiver(payload.name);
      }
      break;
    default:
      break;
  }
});

// Elixir
Popcorn.Wasm.sendEvent("processing_finished", %{id: id})
```

The init process then would look like this:
- send `popcorn_register_receiver` with name payload (`Wasm.register_receiver(pid, name)`)
- send `popcorn_elixir_ready` event (`Wasm.send_elixir_ready()`)

Elixir ready event is sent after registration to not lose any calls but all events will be stored in queue until elixir will be ready.

PR #300, opened 2025-09-12
Enhance Elixir communication with JS

mmietla:
Implements: #128

### DIFF
--- a/examples/eval-in-wasm/lib/eval_in_wasm.ex
+++ b/examples/eval-in-wasm/lib/eval_in_wasm.ex
@@ -11,7 +11,7 @@ defmodule EvalInWasm do
 
   @impl GenServer
   def init(_args) do
-    Wasm.register(@process_name)
+    Wasm.register_default_receiver(self(), @process_name)
     {:ok, nil}
   end
 
--- a/examples/game-of-life/lib/game_of_life.ex
+++ b/examples/game-of-life/lib/game_of_life.ex
@@ -4,8 +4,6 @@ defmodule GameOfLife do
   executing `run/0` when the app is ready.
   """
 
-  alias Popcorn.Wasm
-
   @doc false
   def child_spec(_arg) do
     %{
@@ -20,7 +18,6 @@ defmodule GameOfLife do
   """
   def run() do
     IO.puts("Running...\n")
-    Wasm.register("noop")
     :ignore
   end
 end
--- a/examples/hello-popcorn/lib/hello_popcorn.ex
+++ b/examples/hello-popcorn/lib/hello_popcorn.ex
@@ -9,7 +9,6 @@ defmodule HelloPopcorn do
 
   @impl true
   def init(_init_arg) do
-    Popcorn.Wasm.register(@process_name)
     IO.puts("Hello console!")
 
     Popcorn.Wasm.run_js("""
--- a/examples/iex-wasm/lib/iex_wasm.ex
+++ b/examples/iex-wasm/lib/iex_wasm.ex
@@ -13,7 +13,7 @@ defmodule IexWasm do
 
   @impl GenServer
   def init(_args) do
-    Wasm.register(@process_name)
+    Wasm.register_default_receiver(self(), @process_name)
     type = :elixir
     Shell.start_link(type: type)
     :ok = ExTTY.send_text(:"#{type}_tty", ":ok\n")
--- a/language-tour/elixir_tour/lib/elixir_tour.ex
+++ b/language-tour/elixir_tour/lib/elixir_tour.ex
@@ -21,7 +21,7 @@ defmodule ElixirTour do
 
   @impl GenServer
   def init(_init_arg) do
-    Wasm.register(@process_name)
+    Wasm.register_default_receiver(self(), @process_name)
     :application.set_env(:elixir, :ansi_enabled, false)
     {:ok, %{editor_order: [], bindings: %{}}}
   end
--- a/local-live-view/lib/local_live_view/dispatcher.ex
+++ b/local-live-view/lib/local_live_view/dispatcher.ex
@@ -21,7 +21,7 @@ defmodule LocalLiveView.Dispatcher do
 
   @impl true
   def init(_init_arg) do
-    Popcorn.Wasm.register(@process_name)
+    Wasm.register_default_receiver(self(), @process_name)
     {:ok, %{views: []}}
   end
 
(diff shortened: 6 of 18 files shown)


lib/popcorn/api/wasm.ex
```diff
@@ -240,16 +240,16 @@ defmodule Popcorn.Wasm do
...
-    ({ wasm, args }) => {
-      wasm.onElixirReady?.(args.main);
-    }
-    """
-    |> run_js(%{main: to_string(main_process_name)})
-
+    register_receiver(to_string(main_process_name))
+    send_elixir_ready()
     :ok
   end
```
jgonet: IMO API should be something like this:
```elixir
def main() do
  register(self(), :main)
  Wasm.register_receiver(:main)
  # Wasm.set_default_receiver(:main) # when needing to change default receiver elsewhere
  Wasm.send_elixir_ready(:main) # when registering, you can optionally add process name so we call `Wasm.set_default_receiver` for you
end
```

cc: @bblaszkow06, @mat-hek

priv/static-template/wasm/popcorn.js
```diff
@@ -272,10 +301,60 @@ export class Popcorn {
...
-    this._trace("Main: onInit, main process: ", initProcess);
-    this._mountPromise();
-    this._initProcess = initProcess;
+  _onEvent({ eventName, payload }) {
+    this._trace("Main: onEvent: ", { eventName, payload });
+
+    if (eventName === "popcorn_register_receiver") {
+      if (this.defaultReceiver === null) {
+        this.setDefaultReceiver(payload.name);
+      }
```
jgonet: This behavior is unpredictable from developer perspective. They didn't ask to set default receiver _sometimes_ if they register any receiver.

priv/static-template/wasm/popcorn.js
```diff
@@ -272,10 +301,60 @@ export class Popcorn {
...
+
+      if (this._mountPromise !== null) {
+        this._mountPromise();
+        this._mountPromise = null;
+      }
```
jgonet: Same here.

priv/static-template/wasm/popcorn.js
```diff
@@ -272,10 +301,60 @@ export class Popcorn {
...
+      return;
+    }
+
+    if (eventName === "popcorn_elixir_ready") {
+      this._elixirReady = true;
```
jgonet: This is determined by mount promise resolution, we don't need that state.

priv/static-template/wasm/popcorn.js
```diff
@@ -272,10 +301,60 @@ export class Popcorn {
...
+      this._processEventQueue();
+      this._dispatchEvent(eventName, payload);
+      return;
+    }
+
+    if (!this._elixirReady) {
+      this._eventQueue.push({ eventName, payload });
```
jgonet: I don't like unbounded queues. We can at least issue a warning if more than 10 messages are in the queue and Elixir still didn't finish initializing.

priv/static-template/wasm/popcorn.js
```diff
@@ -272,10 +301,60 @@ export class Popcorn {
...
+
+    this._dispatchEvent(eventName, payload);
+  }
+
+  _processEventQueue() {
+    this._trace(
+      "Main: processing event queue, length: ",
+      this._eventQueue.length
+    );
+    const events = [...this._eventQueue];
```
jgonet: Why copy?

priv/static-template/wasm/popcorn.js
```diff
@@ -272,10 +301,60 @@ export class Popcorn {
...
+
+    for (const { eventName, payload } of events) {
+      this._dispatchEvent(eventName, payload);
+    }
+  }
+
+  _dispatchEvent(eventName, payload) {
+    this._trace("Main: dispatching event: ", { eventName, payload });
+
+    const handlers = this._eventHandlers.get(eventName) || [];
```
jgonet: `??`

priv/static-template/wasm/popcorn_iframe.js
```diff
@@ -66,7 +66,14 @@ export async function initVm() {
     const serialized = Module.serialize(args);
     return origCall(process, serialized);
   };
+  Module["sendEvent"] = (eventName, payload) => {
+    if (eventName === "popcorn_register_receiver" && !Module._vmInitialized) {
```
jgonet: Elixir ready event, not receiver register event.

jgonet (changes requested):

lib/popcorn/api/wasm.ex
mat-hek: I already forgot the context of this, can we talk through it after standup?

bblaszkow06: So, IIRC the meeting, we agreed to:

* Move `send_elixir_ready` to our generated Boot module
* Have explicit declaration of the default receiver

So, the API would be:

```elixir
def main() do
  Process.register(self(), :main)
  Wasm.register_receiver(:main, default: true)
end
```

I would also consider registering the process from the inside of `register_reciever` to reduce the boilerplate, but we'd have to handle the possibility to register via `name: :whatever` passed to `GenServer.start_link`, so we'd need something like:

```elixir
def register_receiver(pid, name, opts \\ []) do
  default? = Keyword.get(opts, :default, false)
  case Process.whereis(name) do
    nil -> Process.register(pid, name)
    ^pin -> :ok
    _other_pid -> raise "Name #{name} already taken"
  end
  register_receiver_js(to_string(name))
end
```

jgonet: #128 description were updated after yesterday's meeting.

mmietla:
Currently, there is a global callback event handler parameter in the `popcorn.init` function. The `elixir_ready` event is dispatched during the initialization process and resolves `mountPromise`. This causes a problem when registering a handler with `popcorn.OnMessage(eventName, callback)`, because the popcorn instance is unavailable, and events dispatched before elixir_ready are not passed through user handlers.


**Option 1: static method Registration**

Register event handlers before any instance exists using static methods

```js
 Popcorn.onMessage("elixir_ready", () => {
     console.log("elixir_ready event");
 });

 Popcorn.onMessage("init_test", (payload) => {
     console.log("Init test:", payload);
 });

 const popcorn = await Popcorn.init({
     debug: true,
     onStdout: console.log,
     onStderr: console.error
 });
```

**Option 2: eventHandlers configuration (modification of global handler)**

Pass handlers directly in the init options

```js
  const popcorn = await Popcorn.init({
      debug: true,
      onStdout: console.log,
      onStderr: console.error,

      eventHandlers: {
          "elixir_ready": () => {
              console.log("🎉 Elixir is ready!");
          },
          "init_test": (payload) => {
              console.log("Init test:", payload);
          },
          ...
      }
  });
  ```

PR #301, opened 2025-09-12
Adjust to the downstreamed AtomVM

mat-hek:


patches/otp/erts/init.erl
```diff
@@ -1,8 +1,11 @@
...
--export([get_arguments/0, get_argument/1]).
+-export([get_arguments/0, get_argument/1, boot/1]).
 
 % Patch reason: this is a mock implementation,
 % sufficient for some use cases not to break
 get_arguments() -> [].
 get_argument(_Arg) -> error.
+
+boot([<<"-s">>, StartupModule]) when is_atom(StartupModule) ->
+    StartupModule:start().
```
bblaszkow06, line 16: I would add patch reason here. ~Is there any documentation on the AtomVM boot process? Or the PR that changed it, that I guess introduced the need for this patch?~
EDIT: I've found the link https://github.com/atomvm/AtomVM/pull/1752
BTW, since we patch it, we should consider moving some of the code from our Boot module

mat-hek: Yup, will add the patch reason. I don't think we can move the code from the boot module though, since it's generated per app, and the patches are static

test/popcorn/eval_test.exs
```diff
@@ -607,8 +607,9 @@ defmodule Popcorn.EvalTest do
     |> AtomVM.assert_result([2, 3, 4, 5])
   end
 
+  @tag :skip
```
mat-hek, line 610: We need to skip this because https://github.com/atomvm/AtomVM/pull/1804 isn't included in https://github.com/software-mansion-labs/FissionVM/pull/151. I'm not sure whether to include it - but if so, maybe in a separate PR

PR #305, opened 2025-09-16
Mmietla/lang tour fixes

mmietla:
Implements feedback from: #296 , #292 

Also:
- Fixed navigation positioning on mobile devices
- Added a Cancel button for stopping code execution

### DIFF
--- a/misc/language-tour-guide/src/context/popcorn/PopcornProvider.tsx
+++ b/misc/language-tour-guide/src/context/popcorn/PopcornProvider.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, type ReactNode } from "react";
+import { useState, useEffect, type ReactNode, useCallback } from "react";
 import { PopcornContext, type Popcorn, type PopcornContextValue } from ".";
 import { useCodeEditorStore } from "../../store/codeEditor";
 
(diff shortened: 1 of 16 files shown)


misc/language-tour-guide/src/context/popcorn/PopcornProvider.tsx
```diff
@@ -14,24 +14,35 @@ export const PopcornProvider = ({
...
-        console.error("Failed to initialize Popcorn:", error);
+        instance.deinit();
+        initializePopcorn();
+      } catch (e) {
+        console.error("Error during Popcorn reinitialization:", e);
       }
     }
+  }, [instance, initializePopcorn]);
 
+  useEffect(() => {
```
jgonet, line 45: Why we don't have `instance` in the deps?

misc/language-tour-guide/src/utils/storage/index.ts
```diff
@@ -0,0 +1,9 @@
+export function hashPathname(pathname: string): string {
+  let hash = 0;
+  for (let i = 0; i < pathname.length; i++) {
+    const char = pathname.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash = hash & hash;
+  }
+  return Math.abs(hash).toString(36);
+}
```
jgonet, line 9: Good enough for now but I'd consider using some standardized non-crypto hash function like xxhash or murmur.

misc/language-tour-guide/src/context/popcorn/PopcornProvider.tsx
mmietla: We want to initialize the popcorn instance when the component mounts. If the instance is included in the dependency array, it will cause an infinite re-render loop.

PR #309, opened 2025-09-22
Handle printing to standard_error from io:put_chars

bblaszkow06:
Depends on https://github.com/software-mansion-labs/FissionVM/pull/152

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,8 +1,5 @@
 import Config
 
-config :popcorn,
-  runtime: [{:path, Path.join(__DIR__, "../popcorn_runtime_source/artifacts/$target")}]
-
 if File.exists?("#{__DIR__}/config.secret.exs") do
   # Put the following configuration in config.secret.exs:
   # config :popcorn, runtime_path: ...
--- a/lib/popcorn/config.ex
+++ b/lib/popcorn/config.ex
@@ -5,7 +5,7 @@ defmodule Popcorn.Config do
     out_dir: nil,
     add_tracing: false,
     runtime: [
-      {:url, "https://github.com/software-mansion-labs/FissionVM/releases/download/popcorn-0.1/",
+      {:url, "https://github.com/software-mansion-labs/FissionVM/releases/latest/download/",
        target: :wasm},
       {:path, "popcorn_runtime_source/artifacts/$target"}
     ]


mat-hek (approved):

config/config.exs
```diff
@@ -1,5 +1,8 @@
 import Config
 
+config :popcorn,
+  runtime: [{:path, Path.join(__DIR__, "../popcorn_runtime_source/artifacts/$target")}]
```
mat-hek: I don't think it's the best approach, as it'll require each project using popcorn master to have it configured and FissionVM built from source for `put_chars` to work. I think we should release FissionVM latest instead. And change `config.ex` to use latest.

bblaszkow06: The fact that new FissionVM version should be released doesn't affect this PR - this line should be here for tests to be running the same FissionVM version, so I strongly disagree the release should be done *instead*.

A permanent solution could go different way and download the artifact for unix as well, but it's more complicated and out of scope for this PR

mat-hek: If we release FissionVM latest and change the `config.exs` back, and update `config.ex`, the tests will pass. I don't see how downloading artifacts for unix would fix this situation, since it's all about artifacts for wasm. If we merge this as is, then popcorn will stop working on master when used as described in the readme. Unless I'm missing something.

bblaszkow06: > If we release FissionVM latest and change the config.exs back, and update config.ex, the tests will pass.

This is just fighting the symptoms, not fixing the underlying cause.

> I don't see how downloading artifacts for unix would fix this situation, since it's all about artifacts for wasm.

It would make it consistent between wasm and unix tests, as now they are using different versions of FissionVM. Only when current master is the same as last release these versions match, otherwise it's a mess.

> If we merge this as is, then popcorn will stop working on master when used as described in the readme. Unless I'm missing something.

What do you mean by "stop working"? This is a library, so the main way it should be used is as a dependency, which remains untouched. Running tests for WASM would require running `mix popcorn.build_runtime --target wasm`. I can add that to README if needed

mat-hek: > This is a library, so the main way it should be used is as a dependency, which remains untouched

If you add a github or path dependency, like we do in examples, it won't work

> Running tests for WASM would require running `mix popcorn.build_runtime --target wasm`

Not only that, just using popcorn as a dependency would require building WASM from source. We can change that, but:
- We need to change the default in `config.ex`, not `config.exs`
- We need to update readme in both popcorn and examples
- We need to either remove the readme from the docs and have a separate homepage for docs, or change the readme each time we publish the docs, as using the hex.pm popcorn and popcorn master would differ
- It would require everyone who clones the repo and runs our examples to build FissionVM from source
- I think we should align it with whatever @jgonet is changing in the build system

I agree it's bad that we don't use unix and wasm the same way. It's because the priority is to make it simple to use wasm in the first place, even from master. Maybe we should change that and prioritise development more.

bblaszkow06: Leaving this as a separate issue then.
I've released the new version of Fission and updated the default for config

mat-hek (approved):

Issue #311, opened 2025-09-25
[Lang tour] Default code in the editor doesn't match the description


mat-hek:
Oftentimes, the default content of the code editor is loaded from a wrong chapter, not the one currently displayed in the left pane. It seems the bug was introduced in https://github.com/software-mansion/popcorn/pull/305

https://github.com/user-attachments/assets/bf5df431-dac4-4af9-8539-f24442b5319d

PR #314, opened 2025-09-30
Lang tour wrong content loading fix

mmietla:
fix: #311

### DIFF
--- a/misc/language-tour-guide/src/store/codeEditor.ts
+++ b/misc/language-tour-guide/src/store/codeEditor.ts
@@ -23,8 +23,9 @@ export const useCodeEditorStore = create<CodeEditorStore>((set, get) => ({
 
   setCode: (code: string) =>
     set((state) => {
-      if (state.pathHash) {
-        localStorage.setItem(`code-${state.pathHash}`, code);
+      const pathHashForStorage = state.pathHash;
+      if (pathHashForStorage) {
+        localStorage.setItem(`code-${pathHashForStorage}`, code);
       }
 
       return { code: code };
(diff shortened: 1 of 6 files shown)


misc/language-tour-guide/src/store/codeEditor.ts
```diff
@@ -23,8 +23,9 @@ export const useCodeEditorStore = create<CodeEditorStore>((set, get) => ({
 
   setCode: (code: string) =>
     set((state) => {
-      if (state.pathHash) {
-        localStorage.setItem(`code-${state.pathHash}`, code);
+      const pathHashForStorage = state.pathHash;
+      if (pathHashForStorage) {
+        localStorage.setItem(`code-${pathHashForStorage}`, code);
```
jgonet, line 28: Do we ever clear this in any way? Changing hash function invalidates all previous entries and changing content does it as well (?)

jgonet (approved):

misc/language-tour-guide/src/store/codeEditor.ts
mmietla: We don't clear this local storage, and changing the content does not invalidate previous entries.

jgonet: Hm, we may consider it then.

mmietla: I am not sure when we can clear this saved code.

mmietla: I thought about storing the defaultCode hash and comparing it with the actual value. When they differ, we would clear the entry. The second idea is to use a hash of `pathname + defaultCode` as the key, but this leaves the old localStorage entries uncleared. 
cc @jgonet

jgonet: You can file an issue for now, it's not critical to fix that immediately.

PR #316, opened 2025-10-01
Improve editing content in language tour

mmietla:
Replaced usage of the defaultCode variable in MDX files with explicit code blocks wrapped in the <EditorCode> component.

Example change:

```
<EditorCode>
```elixir
-- code goes here --
.```
</EditorCode> 
```

### DIFF
--- a/misc/language-tour-guide/src/plugins/remarkCollectCode.ts
+++ b/misc/language-tour-guide/src/plugins/remarkCollectCode.ts
@@ -1,4 +1,4 @@
-import { visit, SKIP } from "unist-util-visit";
+import { visit, SKIP, type Test } from "unist-util-visit";
 import type { Node, Parent } from "unist";
 
 interface MdxJsxElement extends Node {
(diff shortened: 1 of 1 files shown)


misc/language-tour-guide/src/plugins/remarkCollectCode.ts
```diff
@@ -0,0 +1,90 @@
...
+                }
+              ]
+            },
+            specifiers: [],
+            source: null
+          }
+        ]
+      }
+    }
+  } as Node;
```
jgonet: Why do you need to cast it? i.e. why the structure needs `value: ""`?

misc/language-tour-guide/src/plugins/remarkCollectCode.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function remarkCollectCode() {
+  return (tree: Parent) => {
+    const codeBlocks: string[] = [];
+
+    visit(tree, (node: Node, index, parent) => {
+      if (
+        node.type === "mdxJsxFlowElement" &&
+        (node as MdxJsxElement).name === "EditorCode"
```
bblaszkow06: Maybe you could define [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

misc/language-tour-guide/src/plugins/remarkCollectCode.ts
```diff
@@ -0,0 +1,90 @@
...
+      ) {
+        if (parent !== undefined && parent && index !== undefined) {
+          (parent as Parent).children.splice(index, 1);
+        }
```
bblaszkow06: If you'd properly type the callback (i.e. `(node: Node, index?: number, parent: Parent?)` you can simplify that

```suggestion
        if (index !== undefined) {
          parent?.children.splice(index, 1);
        }
```

misc/language-tour-guide/src/plugins/remarkCollectCode.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        const codeBlockNode = node as MdxJsxElement;
```
bblaszkow06: With type predicate this should not be needed

misc/language-tour-guide/src/plugins/remarkCollectCode.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        if (!codeBlockNode.children || codeBlockNode.children?.length === 0) {
+          return [SKIP, index];
+        }
+
+        const code = codeBlockNode.children[0] as CodeNode;
```
bblaszkow06: I think this should be a type check, not a simple assertion

misc/language-tour-guide/src/plugins/remarkCollectCode.ts
bblaszkow06: Why do you need that cast?

mmietla: `tree.children` is of type `Node`, and the property `value` is in the node of type `mdxjsEsm`

mmietla: > `tree.children` is of type `Node`, and the property `value` is in the node of type `mdxjsEsm`

misc/language-tour-guide/src/plugins/remarkCollectCode.ts
```diff
@@ -0,0 +1,103 @@
...
+    const codeBlocks: string[] = [];
+
+    visit<Parent, Test>(tree, (node, index, parent) => {
+      const isCodeBlock = isMdxFlowElement(node) && node.name === "EditorCode";
+      if (!isCodeBlock) {
+        return undefined;
+      }
+
+      const canDropNodeFroParent = index !== undefined && parent !== undefined;
+      if (canDropNodeFroParent) {
```
bblaszkow06, line 43: typo:
```suggestion
      const canDropNodeFromParent = index !== undefined && parent !== undefined;
      if (canDropNodeFromParent) {
```

bblaszkow06 (approved):
LGTM!

jgonet (approved):

PR #318, opened 2025-10-01
[Lang tour] chapter 1 basic types and operations

mat-hek:


misc/language-tour-guide/src/content/1-basic-types-and-operations/atoms.mdx
```diff
@@ -0,0 +1,15 @@
+export const frontmatter = { order: 17 };
+
+# Atoms
+
+An atom is a constant whose value is its own name. Some other languages call these symbols.
+
+`true`, `false` and `nil` are also atoms - Elixir allows to skip the colon for them.
+
+export const defaultCode = `IO.inspect :apple == :apple, label: "atoms with the same name are equal"
```
bblaszkow06: Since you're using IO.inspect with label so much, I think you should rebase onto main and mention that parameter in introduction.

Also, I think you should use parens in those IO.inspect calls - formatter adds them, and it is never mentioned that you may call functions like that

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-arithmetic.mdx
bblaszkow06, line 1: You have omitted a pages on Integers and Floats.
1. I think it would make sense to have them in contents list
2. The info that ints have arbitrary precision is missing
3. `is_number` is missing
4. `ceil` is missing

misc/language-tour-guide/src/content/1-basic-types-and-operations/booleans-and-nil.mdx
bblaszkow06, line 1: You have omitted the rule of thumb suggestion from Elixir guide:
> As a rule of thumb, use and, or and not when you are expecting booleans. If any of the arguments are non-boolean, use &&, || and !.

misc/language-tour-guide/src/content/1-basic-types-and-operations/booleans-and-nil.mdx
```diff
@@ -0,0 +1,28 @@
+export const frontmatter = { order: 16 };
+
+# Booleans and nil
+
+Elixir supports `true` and `false` as booleans, and provides logical `and`, `or` and `not` operators to work on them.
+
+Elixir also provides the concept of `nil`, to indicate the absence of a value, and a set of logical operators that also manipulate `nil`: `||/2`, `&&/2`, and `!/1`. For these operators, `false` and `nil` are considered "falsy", all other values are considered "truthy".
```
bblaszkow06: It is important to show these are short-circuit operators and raise in untested branch won't be triggered

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-arithmetic.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+# Basic Arithmetic
+
+Let's run the snippet and play with basic arithmetic operations. Notice, that `10 / 2` returned `5.0`. Division in Elixir always returns a float. For integer division, use `rem`. For float to integer conversion, you can use `round` or `trunc`. There are also `is_integer` and `is_float` to check the type of a number.
+
+<EditorCode>
+```elixir
+IO.inspect 1 + 2, label: "sum 1 and 2"
+IO.inspect 5 * 6, label: "multiply 5 by 6"
+# IO.inspect 10 / 2, label: :division
```
bblaszkow06: You can uncomment that since the FissionVM was updated

misc/language-tour-guide/src/content/1-basic-types-and-operations/strings2.mdx
```diff
@@ -0,0 +1,14 @@
+export const frontmatter = { order: 19 };
+
+# Strings
+
+Strings in Elixir are represented internally by contiguous sequences of bytes, known as binaries. Thus, we can use functions such as `byte_size` on strings. Functions specific to strings are in the `String` module.
```
bblaszkow06, line 5: You have omitted the UTF-8 and the difference between `byte_size` and `String.length`. was that intentional?

misc/language-tour-guide/src/content/1-basic-types-and-operations/structural-comparison.mdx
```diff
@@ -0,0 +1,17 @@
+export const frontmatter = { order: 20 };
+
+# Structural comparison
+
+Elixir also provides `==`, `!=`, `<=`, `>=`, `<` and `>` as comparison operators. We can compare numbers, atoms, strings, booleans, etc. Integers and floats compare the same if they have the same value, but you can use the strict comparison operator `===` and `!==` if you want to distinguish between integers and floats.
```
bblaszkow06, line 5: Important thing here is the order of terms when comparing and caveats that come from that

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-arithmetic.mdx
mat-hek: I'm not sure if we should include all that. The intent is to give a bird's-eye view on the language, not to include particular details. The Elixir's intro doesn't mention all that and I think we should follow it in this aspect

mat-hek:
@bblaszkow06 there were some force pushes, but relevant changes are in the last commit

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-arithmetic.mdx
bblaszkow06: > The intent is to give a bird's-eye view on the language, not to include particular details.

Well, that's debatable as we never explicitly stated this. Anyway, separate chapters on Ints and Floats is something that both Gleam tour does, pointing to docs for modules related to each type.

IMO there's enough content to split that into 2 pages and show a complete set of functions.

Nonetheless, the most important thing for me, that is still missing is the int precision. Gleam mentions that it's 64-bit float on JS runtime and unlimited precision on ERTS (https://tour.gleam.run/basics/ints/), we can do something similar and mention that ints larger than 64-bits won't be supported in your browser (you'll get `:overflow` error)

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-arithmetic.mdx
```diff
@@ -0,0 +1,22 @@
...
+IO.inspect 1 + 2, label: "sum 1 and 2"
+IO.inspect 5 * 6, label: "multiply 5 by 6"
+IO.inspect 10 / 2, label: :division
+IO.inspect div(10, 2), label: "integer division of 10 and 2"
+IO.inspect rem(10, 3), label: "remainder of integer division of 10 and 3"
+
+# Let's change 'round' to 'trunc'. How does the result change?
+IO.inspect round(3.58), label: "round 3.58"
+
+# Let's try changing 'is_integer' to 'is_float'
```
bblaszkow06: nit: Let's not repeat "Let's"
```suggestion
# Try changing 'is_integer' to 'is_float'
```

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-arithmetic.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+# Basic Arithmetic
+
+Let's run the snippet and play with basic arithmetic operations. Notice, that `10 / 2` returned `5.0`. Division in Elixir always returns a float. For integer division, use `rem`. For float to integer conversion, you can use `round` or `trunc`. There are also `is_integer` and `is_float` to check the type of a number.
+
+<EditorCode>
+```elixir
+IO.inspect 1 + 2, label: "sum 1 and 2"
+IO.inspect 5 * 6, label: "multiply 5 by 6"
+IO.inspect 10 / 2, label: :division
```
bblaszkow06: nit: Keep consistency

```suggestion
IO.inspect 10 / 2, label: "division"
```

misc/language-tour-guide/src/content/1-basic-types-and-operations/atoms.mdx
```diff
@@ -0,0 +1,20 @@
+export const frontmatter = { order: 5 };
+
+
+# Atoms
+
+An atom is a constant whose value is its own name. Some other languages call these symbols.
+
```
bblaszkow06, line 6: One more thing that might be added - `:ok` and `:error`. Ideally, with some obvious function call that would return either atom (maybe `IO.puts` ?)

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-types.mdx
```diff
@@ -1,41 +1,15 @@
...
-IO.puts("atom is_atom: #{is_atom(atom)}")
-IO.puts("string is_binary: #{is_binary(string)}")
-IO.puts("list is_list: #{is_list(list)}")
-IO.puts("tuple is_tuple: #{is_tuple(tuple)}")
-IO.puts("map is_map: #{is_map(map)}")`
+IO.inspect 42, label: "integer"
+IO.inspect 3.14159, label: "float"
+IO.inspect true, label: "boolean"
+IO.inspect :elixir, label: "atom"
+IO.inspect "Hello, Elixir!", label: "string"
```
bblaszkow06, line 13: A loose thought - what about charlists and sigils? Where do we include that

misc/language-tour-guide/src/content/1-basic-types-and-operations/booleans-and-nil.mdx
bblaszkow06: You sure you don't want to include that in text as well?

misc/language-tour-guide/src/content/1-basic-types-and-operations/structural-comparison.mdx
```diff
@@ -0,0 +1,23 @@
...
+```elixir
+IO.inspect 1 == 1, label: "is 1 equal to 1"
+
+# Let's change 2.0 to 1.0. Are they equal now?
+IO.inspect 1 == 2.0, label: "is 1 equal to 2.0"
+
+# Let's change "bar" to "foo". Are the strings equal if they have the same contents?
+IO.inspect "foo" == "bar", label: "is 'foo' string equal to 'bar' string"
+
+IO.inspect 1 === 1.0, label: "is 1 integer strictly equal to 1.0 float"
```
bblaszkow06, line 20: nit: to show difference suggest removing one `=` from this line

misc/language-tour-guide/src/content/1-basic-types-and-operations/structural-comparison.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+IO.inspect "abc" > 1, label: "is a string greater than a number?"
```
bblaszkow06, line 22: How about including the atoms as well?
```suggestion
IO.inspect "abc" > 1, label: "is a string greater than a number?"

# Try changing the atom name
IO.inspect :infinity > 1, label: "is an atom greater than one?"
```

misc/language-tour-guide/src/content/1-basic-types-and-operations/basic-types.mdx
mat-hek: Elixir's guide has a separate chapter about sigils late in the tutorial: https://hexdocs.pm/elixir/sigils.html. Charlists are mentioned in https://hexdocs.pm/elixir/lists-and-tuples.html and described in https://hexdocs.pm/elixir/binaries-strings-and-charlists.html. I think it makes sense to do the same.

misc/language-tour-guide/src/content/1-basic-types-and-operations/numbers.mdx
```diff
@@ -0,0 +1,25 @@
+export const frontmatter = { order: 3 };
+
+# Numbers
+
+As you can see, numbers in Elixir are either integers or floats. You can use `is_integer`, `is_float` and `is_number` to check what a value is.
+
+Elixir allows expressing numbers in decimal, binary, octal, hexadecimal and exponential notation.
+
+Normally, integers can be arbitrairly big and they never overflow (or underflow). However, the AtomVM runtime which we're using in the browser only supports integers up to 64 bits and overflow raises an error. Floats are always stored on 64 bits and overflow raises an error.
```
bblaszkow06, line 9: How can you overflow a float? I'm not sure how to trigger that

bblaszkow06 (approved):

misc/language-tour-guide/src/content/1-basic-types-and-operations/numbers.mdx
mat-hek:
```elixir
iex> x = 2.0 ** 512
1.3407807929942597e154
iex> x * x
** (ArithmeticError) bad argument in arithmetic expression: 1.3407807929942597e154 * 1.3407807929942597e154
    :erlang.*(1.3407807929942597e154, 1.3407807929942597e154)
    iex:10: (file)
```

Possibly it's because floats on the BEAM don't support infinity 🤔

bblaszkow06: That's right, it overflows the exponent and should become +Inf, but Erlang is not IEEE 754 compliant and doesn't support infinities or NaNs https://stackoverflow.com/a/43383146

Issue #323, opened 2025-10-03
Exception.format not working


mmietla:
Exception.format/2 causes the code to freeze.

Example:

```elixir
case eval(code) do
  {:error, error_message} ->
    IO.puts(inspect(error_message))
    formatted_error = Exception.format(:error, error_message)
    IO.puts(formatted_error)
    {:reject, inspect(error_message), state}

  result ->
    {:resolve, inspect(result), state}
end
```

`error_message` is printed correctly with `IO.puts(inspect(error_message))`. But when calling `Exception.format(:error, error_message)`, the code freezes and `formatted_error` is never printed. This causes a Promise timeout in popcorn.call.

mat-hek:
Calling `Exception.format(:error, %RuntimeError{})` on Unix makes AtomVM hang and consume 100% CPU 🫠

PR #325, opened 2025-10-03
Remove AtomVM stdlib's implementation of the Exception module

mat-hek:
Fixes https://github.com/software-mansion/popcorn/issues/323

### DIFF
--- a/patches/elixir/elixir/function_clause_error.ex
+++ b/patches/elixir/elixir/function_clause_error.ex
@@ -1,0 +1,16 @@
+defmodule FunctionClauseError do
+  # Patch reason: without the patch this module fails to load
+  # and aborts AtomVM with message 'Unexpected operand, expected a literal (37)'
+
+  def message(_exception) do
+    "no function clause matching"
+  end
+
+  def blame(exception, stacktrace) do
+    {exception, stacktrace}
+  end
+
+  def blame(_exception, _inspect_fun, _fun) do
+    ""
+  end
+end
--- a/test/popcorn/eval_test.exs
+++ b/test/popcorn/eval_test.exs
@@ -733,7 +733,8 @@ defmodule Popcorn.EvalTest do
     result =
       quote do
         try do
-          raise "foo"
+          # Trigger FunctionClauseError
+          String.length(:foo)
         rescue
           e -> Exception.format(:error, e, __STACKTRACE__)
         end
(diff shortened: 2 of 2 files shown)


patches/elixir/elixir/exception.ex
```diff
@@ -1,443 +1,6 @@
...
-  def exception?(term)
-  def exception?(%_{__exception__: true}), do: true
-  def exception?(_), do: false
-
-  @doc """
-  Gets the message for an `exception`.
-  """
-  def message(%module{__exception__: true} = exception) do
-    try do
-      # for some reason, the `module.message(exception)` is crashing when invoked with FunctionClauseError
```
bblaszkow06, line 82: Have you verified this is no longer the case?

patches/elixir/elixir/exception.ex
```diff
@@ -1,443 +1,6 @@
...
-  @doc since: "1.5.0"
-  @spec blame(:error, any, stacktrace) :: {t, stacktrace}
-  @spec blame(non_error_kind, payload, stacktrace) :: {payload, stacktrace} when payload: var
-  def blame(kind, error, stacktrace)
-
-  def blame(:error, error, stacktrace) do
-    %module{} = struct = normalize(:error, error, stacktrace)
-
-    # TODO: Code.ensure_loaded?(module) is not supported right now
-    # original code: Code.ensure_loaded?(module) and function_exported?(module, :blame, 2)
```
bblaszkow06, line 194: Again - does it work? IMO it should be tested

mat-hek:
@bblaszkow06 So I went through the blame of `Exception.ex` and the changes you pointed out seem to be the only patches we introduced. `Code.ensure_loaded?` seems to work fine in both Unix and WASM. The problem with the FunctionClauseError exists indeed, and it's because loading this module breaks the AtomVM, I'm not sure why. For now, I patched away this module to an absolute minimum and it doesn't seem to break anything. We'll need to give it a proper look https://github.com/software-mansion/popcorn/issues/347

patches/elixir/elixir/function_clause_error.ex
```diff
@@ -0,0 +1,16 @@
+defmodule FunctionClauseError do
+  # Patch reason: without the patch this module fails to load
+  # and aborts AtomVM with message 'Unexpected operand, expected a literal (37)'
+
+  def message(_exception) do
+    "no function clause matching"
```
bblaszkow06, line 6: Could you print at least some stacktrace info? Without it, it's gonna be really annoying to debug apps using Popcorn, including snippets in lang tour

bblaszkow06 (changes requested):

patches/elixir/elixir/function_clause_error.ex
mat-hek: The stacktrace info is added by the Exception module itself that calls this callback, no need to add it here

Issue #326, opened 2025-10-06
[Lang tour] Displaying errors from evaluated code


mmietla:
Display error in the execution section from the evaluated code

Issue #321, opened 2025-10-02
Clean saved code in local storage on content change


mmietla:
Option 1:
Store the defaultCode hash and compare it with the actual value. If they differ, clear the entry.

Option 2:
Use a hash of pathname + defaultCode as the key. However, this approach leaves old localStorage entries uncleared.

PR #327, opened 2025-10-06
Lang tour handle error and clear local storage

mmietla:
Implements: #326 #321


misc/language-tour-guide/elixir_tour/lib/elixir_tour.ex
```diff
@@ -23,12 +23,16 @@ defmodule ElixirTour do
 
   defp handle_wasm({:wasm_call, ["eval_elixir", code]}, state) do
     try do
-      result = eval(code)
-      {:resolve, inspect(result), state}
+      case eval(code) do
+        {:error, error_message} ->
+          {:reject, inspect(error_message), state}
+
+        result ->
```
jgonet: ⚠️ It's weird we have just value on success and tuple on error. Can we stick to `{:ok, value} | {:error, reason}`?

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -55,18 +57,31 @@ export function Results() {
...
+        setErrorData(null);
+        setResultData(null);
+        setDurationMs(null);
+
         try {
           result = await call(["eval_elixir", code], {
             timeoutMs: 10_000
           });
-        } catch (error) {
+        } catch (error: any) {
```
jgonet: nit: you can use predicates for errors:
```ts
catch (error: unknown) {
  if (isPopcornError(error) {
    // ...
}

function isPopcornError(error: unknown): error is SomeError {
  return Object.hasOwn(error, "error");
}
```

misc/language-tour-guide/src/utils/content/tree-builder.ts
```diff
@@ -6,6 +7,13 @@ export async function getRawMdxTree() {
   const loadedFlat = await Promise.all(modules.map(load));
   loadedFlat.sort(sortPathLength);
 
+  const defaultCodeConcatenated = loadedFlat.reduce((acc: string, entry) => {
+    acc += entry.hashDefaultCode ?? "";
+    return acc;
+  }, "");
```
jgonet: nit: or `(all: string, entry) => all + entry.hashDefaultCode ?? ""`

misc/language-tour-guide/src/utils/storage/index.ts
```diff
@@ -1,8 +1,25 @@
...
+const H64 = XXH.h64(SEED);
 
 export function hashPathname(pathname: string): string {
-  return H.update(pathname).digest().toString(16);
+  return H32.update(pathname).digest().toString(16);
+}
+
+export function hashDefaultCode(code: string): string {
+  return H64.update(code).digest().toString(16);
+}
```
jgonet: Why we need different hash functions here?

misc/language-tour-guide/src/utils/storage/index.ts
```diff
@@ -1,8 +1,25 @@
...
+export function manageDefaultCodeStorage(concatenatedDefaultCodeHash: string) {
+  const hash = hashDefaultCode(concatenatedDefaultCodeHash);
+  const previousHash = localStorage.getItem("defaultCodeHash");
+
+  if (previousHash === hash) {
+    return;
+  }
+
+  localStorage.clear();
+  localStorage.setItem("defaultCodeHash", hash);
```
jgonet: Shouldn't it live somewhere else, near the code that sets the localstorage?

misc/language-tour-guide/src/utils/storage/index.ts
mmietla: I though i helps maintain uniqueness, because it is 64 bit version.

jgonet: I think it doesn't matter, although it doesn't matter to the extend we can just stick to current impl. Maybe small thing: it isn't really specific to both pathnames and default codes, we could consider naming those `hash32` and `hash64`.
And also would be nice to write it in the comment where you use it to document the decision for using 64bit version.

misc/language-tour-guide/src/utils/storage/index.ts
mmietla: I put it here because it handles saving and deleting items from local storage. I want to keep all local storage–related functions in this file. However, putting it in the content-related files might be fine as well, I think.

PR #329, opened 2025-10-06
Add single-line logs when building bundles

jgonet:
Prints all operations in the same line (in interactive terminals) which limits log spam.

## Screencast

Click to play.

[asciicast](https://asciinema.org/a/BfXBCK37hOpT0ncDukTpt4za8)

### DIFF
--- a/lib/popcorn/build.ex
+++ b/lib/popcorn/build.ex
@@ -1,6 +1,8 @@
 defmodule Popcorn.Build do
   @moduledoc false
+  import Popcorn.BuildLogFormatter, only: [with_build_logger: 1]
   require Popcorn.Config
+  require Logger
   alias Popcorn.CoreErlangUtils
 
   @config Popcorn.Config.compile([:add_tracing, :extra_apps])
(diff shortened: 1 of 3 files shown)


lib/popcorn/build_log_formatter.ex
```diff
@@ -0,0 +1,58 @@
...
+    colors: [enabled: true],
+    metadata: [:app_name]
+  ]
+
+  def with_build_logger(f) do
+    original_formatter = Logger.default_formatter()
+    formatter = Logger.default_formatter(@overrides)
+    :logger.update_handler_config(:default, :formatter, formatter)
+    f.()
+    :logger.update_handler_config(:default, :formatter, original_formatter)
```
mat-hek:
```suggestion
    after ->
      :logger.update_handler_config(:default, :formatter, original_formatter)
```

lib/popcorn/build.ex
```diff
@@ -33,54 +35,60 @@ defmodule Popcorn.Build do
...
+        )
+        |> Map.new()
+
+      popcorn_lib_cache = build_popcorn(cache[:popcorn_lib])
+      new_cache = Map.put(new_cache, :popcorn_lib, popcorn_lib_cache)
+
+      File.write!(@cache_path, :erlang.term_to_binary(new_cache))
+
+      :ok
+    end)
```
mat-hek: Personally, I'd either have `with_build_logger(fn -> do_build(opts) end)` or, preferably, split `with_build_logger` into `enable_build_logger` and `disable_build_logger` to avoid the additional indent

lib/popcorn/build.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Popcorn.Build do
   @moduledoc false
+  import Popcorn.BuildLogFormatter, only: [with_build_logger: 1]
```
mat-hek: I don't see a reason to import this

lib/popcorn/build.ex
```diff
@@ -81,6 +86,9 @@ defmodule Popcorn.Build do
     File.write!(@cache_path, :erlang.term_to_binary(new_cache))
 
     :ok
+
+    BuildLogFormatter.disable(original_formatter)
+    Logger.info("Bundling finished")
```
mat-hek:
```suggestion
    after ->
      BuildLogFormatter.disable(original_formatter)
      Logger.info("Bundling finished")
```

lib/popcorn/build_log_formatter.ex
```diff
@@ -0,0 +1,61 @@
+defmodule Popcorn.BuildLogFormatter do
+  @moduledoc """
+  Used in Popcorn build and patch process to keep logs short.
+  """
+  require Logger
+
+  @clear_line "\e[2K\r"
```
bblaszkow06: This is available as `"#{IO.ANSI.clear_line()}\r"`

lib/popcorn/build_log_formatter.ex
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  defp format_control_chars do
+    if interactive_terminal?(), do: {@clear_line, ""}, else: {"", "\n"}
+  end
+
+  defp interactive_terminal? do
+    with {:ok, true} <- :standard_io |> :io.getopts() |> Keyword.fetch(:terminal),
+         true <- IO.ANSI.enabled?(),
```
bblaszkow06: Isn't that somewhat duplicated? Why do we need that manual check if `IO.ANSI.enabled?` is supposed to check more-or-less the same thing?

lib/popcorn/build_log_formatter.ex
jgonet: Thanks!

lib/popcorn/build_log_formatter.ex
jgonet: Ah, it does. Elixir on startup calls `prim_tty` which uses `isatty` libc function underneath.

lib/popcorn/build.ex
jgonet: Can't do that, `original_formatter` wouldn't be defined. If we raise before that line it will kill the process regardless and we don't need to do any cleanup.

lib/popcorn/build.ex
```diff
@@ -298,4 +310,12 @@ defmodule Popcorn.Build do
     File.mkdir!(tmp_dir)
     tmp_dir
   end
+
+  defp to_human_duration(us) do
+    case {us / 1_000_000, us / 1_000, :erlang.float(us)} do
+      {s, _, _} when s > 1.0 -> "#{Float.round(s, 2)}s"
+      {_, ms, _} when ms > 1.0 -> "#{Float.round(ms, 2)}ms"
+      {_, _, us} -> "#{Float.round(us, 2)}us"
+    end
```
mat-hek:
```suggestion
    us = :erlang.float(us)
    ms = us / 1_000
    s = ms / 1_000
    
    cond do
      s > 1.0 -> "#{Float.round(s, 2)}s"
      ms > 1.0 -> "#{Float.round(ms, 2)}ms"
      true -> "#{Float.round(us, 2)}us"
    end
```

lib/popcorn/build.ex
mat-hek: > If we raise before that line it will kill the process regardless and we don't need to do any cleanup

@jgonet well, actually I don't agree, because the formatter is global. BTW, this also means that any logs that are simultainously printed along the build will also be affected by this formatter. Can we make sure that the formatter only affects the logs from the build?

lib/popcorn/build.ex
jgonet: I've written this way originally but found it less legible. Why do you think it's better?

mat-hek: I think pattern matching doesn't bring any value in this case and makes this less readable. If you prefer more concise version, you can do:

```suggestion
    cond do
      (s = us / 1_000_000)  > 1.0 -> "#{Float.round(s, 2)}s"
      (ms = us / 1_000) > 1.0 -> "#{Float.round(ms, 2)}ms"
      true -> "#{Float.round(:erlang.float(us), 2)}us"
    end
```

But you can leave it too

lib/popcorn/build.ex
jgonet: We could add a metadata that is matched in the formatter but I feel like it's overkill. Logs from Elixir compiler (e.g. warnings) cause it to leave a line and resume using `/r` on the next one which I personally find acceptable, we don't lose any logs.

mat-hek: Fair enough

PR #330, opened 2025-10-06
Refactor bundling logic

jgonet:
~⚠️ Merge after #329~

## Old description

This PR changes caching strategy for patches and beams from stdlib to improve build-times and simplify the logic.

The old behavior loaded cache manifest (with structure`Map<app_name, Map<beam_name, hash>>`) and checked if it's loaded correctly and isn't stale compared to patches file set (and hashes). If it's stale, we drop all build directories and start from scratch.
Patch hashes were calculated async, at the time when app was building.

The new behavior works by pre-calculating hashes and extracting set of changes between cache and current files state.
We run build process for changed files. Then, we make sure that all input beams are located in output directory.

### DIFF
--- a/lib/popcorn/artifacts_cache.ex
+++ b/lib/popcorn/artifacts_cache.ex
@@ -13,6 +13,8 @@ defmodule Popcorn.ArtifactsCache do
       ...
     }
   }
+
+  See `Popcorn.Build`.
   """
 
   @app_path Mix.Project.app_path()
--- a/lib/popcorn/artifacts_cache.ex
+++ b/lib/popcorn/artifacts_cache.ex
@@ -1,21 +1,20 @@
 defmodule Popcorn.ArtifactsCache do
-  @moduledoc """
-  Artifacts cache is a map, stored in external term format on disk. Stores input files with hashes for the previous build.
-  Used to skip building parts of input files if they did not change.
+  @moduledoc false
+  # Artifacts cache is a map, stored in external term format on disk. Stores input files with hashes for the previous build.
+  # Used to skip building parts of input files if they did not change.
 
-  The stored map has following structure:
-  %{
-    app1: %{
-      file_path1: hash1,
-      file_path2: hash2
-    },
-    app2: %{
-      ...
-    }
-  }
+  # The stored map has following structure:
+  # %{
+  #   app1: %{
+  #     file_path1: hash1,
+  #     file_path2: hash2
+  #   },
+  #   app2: %{
+  #     ...
+  #   }
+  # }
 
-  See `Popcorn.Build`.
-  """
+  # See `Popcorn.Build`.
 
   @app_path Mix.Project.app_path()
   @cache_path "#{@app_path}/popcorn_patch_cache"
@@ -100,10 +99,7 @@ defmodule Popcorn.ArtifactsCache do
         %{}
 
       {:error, reason} ->
-        raise(File.Error,
-          reason: reason,
-          path: @cache_path
-        )
+        raise File.Error, reason: reason, path: @cache_path
     end
   end
 
@@ -127,10 +123,7 @@ defmodule Popcorn.ArtifactsCache do
         :ok
 
       {:error, reason} ->
-        raise(File.Error,
-          reason: reason,
-          path: @cache_path
-        )
+        raise File.Error, reason: reason, path: @cache_path
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/popcorn/artifacts_cache.ex
```diff
@@ -0,0 +1,121 @@
...
+      file_path1: hash1,
+      file_path2: hash2
+    },
+    app2: %{
+      ...
+    }
+  }
+
+  See `Popcorn.Build`.
+  """
```
bblaszkow06: Do we want't to expose this module and its docs? IMO we should hide it `@moduledoc false`

lib/popcorn/artifacts_cache.ex
```diff
@@ -0,0 +1,121 @@
...
+        :ok
+
+      {:error, :enoent} ->
+        :ok
+
+      {:error, reason} ->
+        raise(File.Error,
+          reason: reason,
+          path: @cache_path
+        )
```
bblaszkow06: nit: why use parenthesis and multiple lines?

```suggestion
        raise File.Error, reason: reason, path: @cache_path
```

lib/popcorn/artifacts_cache.ex
```diff
@@ -0,0 +1,121 @@
...
+        :erlang.binary_to_term(cache)
+
+      {:error, :enoent} ->
+        %{}
+
+      {:error, reason} ->
+        raise(File.Error,
+          reason: reason,
+          path: @cache_path
+        )
```
bblaszkow06:
```suggestion
        raise File.Error, reason: reason, path: @cache_path
```

lib/popcorn/build.ex
```diff
@@ -98,155 +113,103 @@ defmodule Popcorn.Build do
...
-          do_patch(app, name, stdlib_beams_by_name[name], path, out_dir)
-          name
+    none_patched =
+      process_async(patch_paths, fn patch_path ->
+        with_tmp_dir(out_dir, fn tmp_dir ->
+          Logger.info("Compiling #{patch_path}", app_name: app)
+
+          patch_path
+          |> compile_patch(tmp_dir)
+          # credo:disable-for-lines:3 Credo.Check.Refactor.Nesting
```
bblaszkow06: I agree with the linter on this, could you flatten that?
Maybe `with_tmp_dir` could return the function that would then become an argument to the `process_async`

mat-hek (changes requested):
The reason why everything was built from scratch when a patch was removed was to ensure that formerly patched modules are no longer patched. Now, when you run mix test, then remove a patch, for example application_controller.erl and run mix test again, everything works, even though it shouldn't. When you rebuild from scratch, tests fail.

To address that, you need to track which modules are affected by which patches, and replace them with the original modules (or remove if the patch created a whole new module). Or just recompile from scratch if any patch is removed, as we do now.

Note that for Elixir patches, there may be multiple modules patched in a single file. While the file may still be there, one of the modules may be removed. It's a rare case, not covered on main either, but if we want a bulletproof solution, it should be taken into account.

BTW, I don't see any changes to how the AtomVM is downloaded and built, which I thought was the main goal of the refactor. Are there any other PRs I'm missing?

lib/popcorn/artifacts_cache.ex
mat-hek: This still stands IMO. Same for `Popcorn.Build` and `Popcorn.BuildLogFormatter`.

lib/popcorn/build.ex
```diff
@@ -259,56 +248,75 @@ defmodule Popcorn.Build do
...
+          maybe_add_tracing(ast, @config.add_tracing)
+        end)
+
+        name
+      end)
+      |> Enum.empty?()
+
+    some_copied = not none_copied
+    some_copied
   end
```
mat-hek, line 284: This effectively returns `true` whenever `beams` is not empty. Let's just return `:ok` and do the check where the function is called.

lib/popcorn/build.ex
```diff
@@ -201,54 +184,60 @@ defmodule Popcorn.Build do
...
-      try do
-        Logger.info("Compiling #{src}", app_name: app)
-        compile_patch(src, tmp_dir)
-
-        Path.wildcard("#{tmp_dir}/*")
-        |> Enum.map(fn path ->
-          name = Path.basename(path)
-          do_patch(app, name, stdlib_beams_by_name[name], path, out_dir)
-          name
+    none_patched =
```
mat-hek: This will be true whenever `patch_paths` is empty. I have an impression that this code was refactored a few times and there's some unneeded complexity left - please have a look at it once more.

mat-hek (changes requested):
All unsolved comments from Bartek's review should be addressed too

mat-hek (approved):

PR #331, opened 2025-10-08
[LangTour] Pattern matching

bblaszkow06:


misc/language-tour-guide/src/content/3-pattern-matching/match-operator.mdx
```diff
@@ -0,0 +1,25 @@
+export const frontmatter = { order: 1 };
+
+# Match operator
+
+Going through this guide, you have seen the `=` operator a lot, used to assign a value to the variable.
+Turns out, it's not just an assignment! `=` in Elixir is called _the match operator_.
```
mat-hek: Actually we used `=` only a few times so far. We also never described what variables are and that they're immutable. I didn't find a place in the elixir guide where this is discussed, it's only mentioned in https://hexdocs.pm/elixir/recursion.html. I think we should put that somewhere, but it would be best to have something like `for`, to show that variables are actually constants. To have `for`, we need variables tho...

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
```diff
@@ -0,0 +1,92 @@
...
+map = %{"a" => 1, :b => 2, :c => "c"}
+%{"a" => one} = map
+IO.inspect(one, label: "one")
+
+# Enforcing specific value for :c, try changing it and see the result
+%{:b => two, :c => "c"} = map
+IO.inspect(two, label: "two")
+# For atom keys, a special syntax is available
+%{c: cee} = map
+IO.inspect(cee, label: "cee")
```
mat-hek: I'd add an example with an empty map

misc/language-tour-guide/src/content/3-pattern-matching/special-patterns.mdx
```diff
@@ -0,0 +1,69 @@
+export const frontmatter = { order: 3 };
+
+# Special patterns
+
+There some additional constructs you may use when building patterns:
```
mat-hek:
```suggestion
There are some additional constructs you may use when building patterns:
```

misc/language-tour-guide/src/content/3-pattern-matching/match-operator.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+Check out the example - after the `1` is assigned to `x`, it can be used for matching.
+Line `1 = x` is legal and means "match x against pattern 1", that is successful.
```
jgonet: "that is successful" sounds weird, maybe "which succeeds?"

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
```diff
@@ -0,0 +1,92 @@
+export const frontmatter = { order: 2 };
+
+# Pattern Matching
+
+The match operator can also be used to destructure data types by matching against more complex patterns.
+
+The pattern may consist of any number of variables, ignore patterns, literal values, and patterns for specific data types.
```
jgonet: "ignore patterns" aren't explained.

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
mat-hek, line 1: I'd split tuples, lists and maps into separate pages

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
```diff
@@ -0,0 +1,92 @@
...
+```
+
+</EditorCode>
+
+## Lists
+
+List matching works similarly to tuples - you can extract every element of the list.
+There is, however, a special pattern related to how a list is built:
+since it's either an empty list (`[]`) or the so-called "cons" with a value (called "head")
+and a reference to the rest of the list (being itself a list, called "tail"),
```
jgonet: This reads bad, idea:
> List matching works similarly to tuples - you can extract every element of the list.
> In Elixir, lists are built internally by storing value ("head") with reference to the rest of the list ("tail"). You can build the list yourself using `|` operator.
> This allows you to easily extract items from the beginning of list, piece by piece.
```
# internal representation
last = []
three = [3 | last]
two = [2 | three]
all = [1 | two]

[1, 2, 3] == all
# or, inlined
[1, 2, 3] = [1 | [2 | [3 | []]]]
```

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
```diff
@@ -0,0 +1,92 @@
...
+# | is also used for prepending to the list
+IO.inspect([0 | tail], label: "list with new head")
+```
+
+</EditorCode>
+
+## Maps
+
+Patterns for maps can extract the value of any key or ensure that a specified key has a provided value.
+The pattern doesn't have to be exhausting.
```
jgonet: "exhausting" -> "exhaustive"

Also we didn't explain what that means.

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
bblaszkow06: They are - in special patterns. Do you think I should mention that?

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
bblaszkow06: The description seems fine, however the code seems redundant, as it's not pattern matching and there is already a chapter on lists (and building them)

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
jgonet: It should be a link at least, if this text comes before.

misc/language-tour-guide/src/content/3-pattern-matching/pattern-matching.mdx
jgonet: Fine by me, I wanted to capture the connection between operator and avoid "X or A with B..." sentence structure, it's hard to read. Not really insisting on my example word-to-word.

misc/language-tour-guide/src/content/3-pattern-matching/matching-tuples.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+# Uncomment to see the match error
+# {a, b} = {:hello, "world", 42}
+
+{a, b, c} = {:hello, "world", 42}
+IO.inspect(a, label: "a")
+IO.inspect(b, label: "b")
+IO.inspect(c, label: "c")
+
+{:ok, answer} = Map.fetch(%{answer: 42}, :answer)
```
mat-hek, line 26: You can suggest changing answer to an unknown key

misc/language-tour-guide/src/content/3-pattern-matching/matching-tuples.mdx
```diff
@@ -0,0 +1,32 @@
...
+<EditorCode>
+
+```elixir
+{x, y, z} = {:hello, "world", 42}
+IO.inspect(x, label: "x")
+IO.inspect(y, label: "y")
+IO.inspect(z, label: "z")
+
+# Uncomment to see the match error
+# {a, b} = {:hello, "world", 42}
```
mat-hek: For such simple things that are easy to describe, I'd go for asking to change the code, not just uncommenting

misc/language-tour-guide/src/content/3-pattern-matching/matching-maps.mdx
mat-hek, line 1: Since the chapter 2 now introduces only lists and tuples, the user won't know about maps yet... I guess I'll add maps then, it should be possible (I'll just avoid mentioning `Map.update` that requires a lambda)

misc/language-tour-guide/src/content/3-pattern-matching/assignment.mdx
mat-hek, line 1: Actually, we could have this in the first chapter 🤔 WDYT?

misc/language-tour-guide/src/content/3-pattern-matching/matching-tuples.mdx
bblaszkow06: I had that as an example, but with current way the errors are shown, I decided to remove it. Can I assume this will be fixed before the first release?

misc/language-tour-guide/src/content/3-pattern-matching/assignment.mdx
bblaszkow06: It is actually very similar to variables chapter - I think I can merge them

misc/language-tour-guide/src/content/3-pattern-matching/matching-tuples.mdx
mat-hek: Yes, otherwise we'll remove all the examples that throw errors ¯\\_(ツ)\_/¯

PR #336, opened 2025-10-13
[Lang tour] Chapter 2 - lists and tuples

mat-hek:


misc/language-tour-guide/src/content/2-lists-and-tuples/introduction-to-lists-in-elixir.mdx
```diff
@@ -0,0 +1,14 @@
+export const frontmatter = { order: 1 };
+
+# Introduction to Lists in Elixir
+
+In this chapter we will learn two of the most used collection data-types in Elixir: lists and tuples. Let's start with lists. Elixir uses square brackets to specify a list of values. Values can be of any type.
```
bblaszkow06: nit: I'd break the paragraph:

```suggestion
In this chapter we will learn two of the most used collection data-types in Elixir: lists and tuples.

Let's start with lists.
Elixir uses square brackets to specify a list of values.
Values can be of any type.
```

misc/language-tour-guide/src/content/2-lists-and-tuples/prepending-to-list.mdx
```diff
@@ -0,0 +1,14 @@
+export const frontmatter = { order: 3 };
+
+# Prepending to list
+
+Lists are stored in memory as linked lists, meaning that each element in a list holds its value and points to the following element until the end of the list is reached. Therefore, prepending to a list is a fast (constant time) operation, while appending requires traversing entire list. That's why prepending an element is a common operation in Elixir and there's a syntax sugar for it: `[element | list]`
```
bblaszkow06: My markdown linter complains about too long lines, and while I won't enforce its suggested 90 characters, it would be better to split this - it would work better with git and reviewing

```suggestion
Lists are stored in memory as linked lists, meaning that each element in a list holds its value
and points to the following element until the end of the list is reached.
Therefore, prepending to a list is a fast (constant time) operation, while appending requires traversing entire list.
That's why prepending an element is a common operation in Elixir and there's a syntax sugar for it: `[element | list]`
```

misc/language-tour-guide/src/content/2-lists-and-tuples/list-operations-in-elixir.mdx
```diff
@@ -0,0 +1,16 @@
+export const frontmatter = { order: 2 };
+
+# List Operations in Elixir
+
+Two lists can be concatenated or subtracted using the `++` and `--` operators respectively. List operators never modify the existing list. Concatenating to or removing elements from a list returns a new list. We say that Elixir data structures are *immutable*. One advantage of immutability is that it leads to clearer code. You can freely pass the data around with the guarantee no one will mutate it in memory - only transform it.
```
bblaszkow06: paragraph split:
```suggestion
Two lists can be concatenated or subtracted using the `++` and `--` operators respectively.
List operators never modify the existing list.
Concatenating to or removing elements from a list returns a new list.

We say that Elixir data structures are *immutable*.
One advantage of immutability is that it leads to clearer code.
You can freely pass the data around with the guarantee no one will mutate it in memory - only transform it.
```

misc/language-tour-guide/src/content/2-lists-and-tuples/introduction-to-lists-in-elixir.mdx
bblaszkow06: > Values can be of any type

I would extend the text of the guide with mention that you can mix them (the getting started shows an example of mixed list right away, if we have a task to change it, I would make sure it is clear).

I'd go with something like

> Values can be of any type, so you can mix different types within one list

bblaszkow06: This is a 1:1 copy from getting started. Doesn't that require some mention/copyright notice or something?

misc/language-tour-guide/src/content/2-lists-and-tuples/prepending-to-list.mdx
bblaszkow06: How is `|` a syntactic sugar? It's THE syntax for it, there's no other syntax that this operator simplifies

bblaszkow06: I'm missing some contents of [`List` module docs](https://hexdocs.pm/elixir/1.17.3/List.html).
Namely, the details of list structure (`[1 | [2 | [3 | []]]]`)

misc/language-tour-guide/src/content/2-lists-and-tuples/error-handling-and-charlists.mdx
```diff
@@ -0,0 +1,13 @@
+export const frontmatter = { order: 5 };
+
+# Error Handling and Charlists
```
bblaszkow06: I don't see any error handling 🤔

misc/language-tour-guide/src/content/2-lists-and-tuples/charlists.mdx
```diff
@@ -0,0 +1,13 @@
+export const frontmatter = { order: 5 };
+
+# Error Handling and Charlists
+
+When Elixir sees a list of printable ASCII numbers, it will print that as a charlist (a list of characters). That's for better interoperability with Erlang, where strings are represented as charlists instead of binaries.
```
bblaszkow06, line 5: Maybe it would be worth mentioning the `~c"charlist"` construct

misc/language-tour-guide/src/content/2-lists-and-tuples/tuples-in-elixir.mdx
```diff
@@ -0,0 +1,13 @@
+export const frontmatter = { order: 6 };
+
+# Tuples in Elixir
+
+Elixir uses curly brackets to define tuples. Like lists, tuples can hold any value. Tuples store elements contiguously in memory which means accessing a tuple element by index or getting the tuple size is a fast operation.
```
bblaszkow06:
```suggestion
Elixir uses curly brackets to define tuples.
Like lists, tuples can hold values of any type.

Tuples store elements contiguously in memory
which means accessing a tuple element by index
or getting the tuple size is a fast operation.
```

misc/language-tour-guide/src/content/2-lists-and-tuples/updating-a-tuple.mdx
```diff
@@ -0,0 +1,16 @@
+export const frontmatter = { order: 7 };
+
+# Updating a tuple
+
+It is possible to put an element at a particular index in a tuple with `put_elem`. Notice that `put_elem` returned a new tuple - that's because tuples, like lists, are also immutable. It means that changing a tuple is expensive - a whole new tuple has to be created in memory.
```
bblaszkow06: ("returned" makes more sense under evaluated code block)
```suggestion
It is possible to put an element at a particular index in a tuple with `put_elem`.
Notice that `put_elem` returns a new tuple -
that's because tuples, like lists, are also immutable.
It means that changing a tuple is expensive -
a whole new tuple has to be created in memory.
```

misc/language-tour-guide/src/content/2-lists-and-tuples/updating-a-tuple.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+Note, however, that the elements themselves are not copied. When you update a tuple, all entries are shared between the old and the new tuple, except for the entry that has been replaced. This rule applies to most data structures in Elixir. This reduces the amount of memory allocation the language needs to perform and is only possible thanks to the immutable semantics of the language.
```
bblaszkow06: "This" is repeated. BTW that could be upstreamed to the getting started 😛 
```suggestion
Note, however, that the elements themselves are not copied.
When you update a tuple, all entries are shared between the old and the new tuple,
except for the entry that has been replaced.
This rule applies to most data structures in Elixir.
Such approach reduces the amount of memory allocation the language needs to perform
and is only possible thanks to the immutable semantics of the language.
```

misc/language-tour-guide/src/content/2-lists-and-tuples/updating-a-tuple.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+<EditorCode>
+```elixir
+tuple = {:ok, "hello"}
+
+# Let's print the 'tuple' variable after the update - did it change?
+IO.inspect put_elem(tuple, 1, "world"), label: "Updated tuple"
```
bblaszkow06, line 19: You may also suggest using out-of-bound index

misc/language-tour-guide/src/content/2-lists-and-tuples/prepending-to-list.mdx
mat-hek: I meant `[1] ++ [2, 3, 4]`. One is concatenation and the other is prepending, but under the hood they are the same thing. And you don't have a counterpart for appending. But I can avoid calling it syntax sugar if you consider it misleading

misc/language-tour-guide/src/content/2-lists-and-tuples/updating-a-tuple.mdx
mat-hek: It's mentioned in getting started, but separately, in the last paragraph

misc/language-tour-guide/src/content/2-lists-and-tuples/introduction-to-lists-in-elixir.mdx
mat-hek: There's even more copy-pasted stuff than this, so definitely a copyright notice is required. I'm wondering if it's enough to put this on the first page

misc/language-tour-guide/src/content/2-lists-and-tuples/prepending-to-list.mdx
mat-hek: This makes editing exceptionally annoying, as I have to manually add and remove newlines. We can agree on a given line length if there's a way to automatically format the document, but I'm not sure it's worth the effort, since the paragraphs are not that long.

misc/language-tour-guide/src/content/2-lists-and-tuples/introduction-to-lists-in-elixir.mdx
bblaszkow06: No idea. Anyway, that's something to solve outside of this PR

mat-hek: I asked Jose about it and he said

> i don't care at all
> you just need to mention the license on the source code

bblaszkow06 (approved):

PR #339, opened 2025-10-14
[Lang tour] Constructs chapter

mat-hek:


bblaszkow06:
Something is wrong with the naming in `misc/language-tour-guide/src/content/4-constructs/case,-cond,-and-if.mdx`, decide if you want dashes or commas 😛 It even breaks the new GH layout, I can't see the diff - had to report that

mat-hek:
@bblaszkow06 done. BTW, the new GH layout is quite unstable for me, so I'm using the old one :P

misc/language-tour-guide/src/content/4-constructs/case-cond-and-if.mdx
```diff
@@ -0,0 +1,26 @@
+export const frontmatter = { order: 1 };
+
+# case, cond, and if
+
+In this chapter, we will learn about the `case`, `cond`, and `if` control flow structures.
+
+`case` allows you to compare a value against many patterns until there is a matching one. You can use `_` as the last, catch-all pattern, that would match any value. Without it, if no pattern matches, an error is raised.
```
bblaszkow06: Technically, this page is mostly about case, so IMHO you should rename it to `case.mdx` and either extract the first paragraph to `intro.mdx` or keep it and add a H2 for `case`

misc/language-tour-guide/src/content/4-constructs/pin-operator.mdx
bblaszkow06, line 1: This duplicates the paragraph from special patterns in pattern matching. Maybe you could rephrase it a bit and point out that whatever was valid for patterns on the left side of match operator works in case as well

misc/language-tour-guide/src/content/4-constructs/guards.mdx
```diff
@@ -0,0 +1,21 @@
+export const frontmatter = { order: 3 };
+
+# Guards
+
+Case also allows extra conditions to be specified via guards. Guards are expressions that can be added to a pattern using the `when` keyword. In the example code, the first clause will only match when `x` is positive.
+
+In guards, you can perform arithmetic expressions, comparisons, type checking (`is_integer`, `is_atom` etc) and call several other functions. They are all listed in the documentation for the [Kernel](https://hexdocs.pm/elixir/Kernel.html) module, in the `Guards` section of sidebar. You can also consult the complete [Patterns and Guards reference](https://hexdocs.pm/elixir/patterns-and-guards.html#guards) for in-depth documentation.
```
bblaszkow06: I'd add one more introductory sentence - something like:

```suggestion
Not any code can be put as a guard.
In guards, you can perform arithmetic expressions, comparisons, type checking (`is_integer`, `is_atom` etc) and call several other functions. They are all listed in the documentation for the [Kernel](https://hexdocs.pm/elixir/Kernel.html) module, in the `Guards` section of sidebar. You can also consult the complete [Patterns and Guards reference](https://hexdocs.pm/elixir/patterns-and-guards.html#guards) for in-depth documentation.
```

PR #351, opened 2025-10-20
[Lang Tour] Refactor error handling

mmietla:
- Refactor error handling in elixir_tour use Exception.format for better error message formatting

image

image


mat-hek (approved):
Eventually we should do `Exception.format(:error, error, __STACKTRACE__)`, but the stacktraces are currently meaningless anyway ¯\\_(ツ)\_/¯

PR #355, opened 2025-10-21
Implement binary:matches as a patch

mat-hek:
`binary:matches` is needed by `Exception.format_snippet` that is called when printing errors, `TokenMissingError` in particular


test/popcorn/eval_test.exs
```diff
@@ -744,4 +744,12 @@ defmodule Popcorn.EvalTest do
 
     assert "** (FunctionClauseError)" <> _rest = result
   end
+
+  async_test "binary:matches" do
```
jgonet:
```suggestion
  async_test "binary:matches", %{tmp_dir: dir} do
```

PR #362, opened 2025-10-22
Lang tour layout adjustments

mmietla:
image


mat-hek (commented):
Looks great! The introduction page says that the output is printed out, let's remove that. Also, I think that this black line next to the output is no longer needed, since there's nothing else there. Let's remove it unless you like it ;)

PR #371, opened 2025-10-24
[iex_wasm] Change prompt prefix from 'avm_iex' to 'local_iex'

mat-hek:
I know that this can be run on Unix as well, but IMO it's more important to make it clearer on the web that it runs in the browser. I also thought about `wasm_iex` or `local_iex`.

### DIFF
--- a/patches/elixir/iex/server.ex
+++ b/patches/elixir/iex/server.ex
@@ -10,7 +10,7 @@ defmodule IEx.Server do
     #     {:popcorn_module.prompt_mode(status, :default), :popcorn_module.default_prefix(status, prefix)}
     #   end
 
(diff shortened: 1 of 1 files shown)


patches/elixir/iex/server.ex
```diff
@@ -10,7 +10,7 @@ defmodule IEx.Server do
     #     {:popcorn_module.prompt_mode(status, :default), :popcorn_module.default_prefix(status, prefix)}
     #   end
 
-    prompt = "avm_iex(#{to_string(counter)})>"
+    prompt = "browser_iex(#{to_string(counter)})>"
```
mat-hek:
```suggestion
    prompt = "local_iex(#{to_string(counter)})>"
```

Issue #368, opened 2025-10-24
[Lang tour] UI/UX community feedback


mat-hek:
@mmietla here's some feedback I found on forum, for you to decide what we should address

- [x] Scroll to the top when clicking “next” - [post](https://elixirforum.com/t/elixir-language-tour-learn-elixir-in-your-browser-with-popcorn-interactive-tutorial/73043/15?u=mat-hek)
- [x] Per page execution history - [post](https://elixirforum.com/t/elixir-language-tour-learn-elixir-in-your-browser-with-popcorn-interactive-tutorial/73043/15?u=mat-hek)
- [x] Make the divider between the instructions on the left and the code/results on the right drag-able - [post](https://elixirforum.com/t/elixir-language-tour-learn-elixir-in-your-browser-with-popcorn-interactive-tutorial/73043/14?u=mat-hek)

PR #372, opened 2025-10-27
[Lang tour] feedback changes

mmietla:
Implements: #368 

https://github.com/user-attachments/assets/f689e165-27d1-4e2f-bca6-b6cebbb4aabd

https://github.com/user-attachments/assets/65432b47-f84c-4309-8c44-caf36f99129a

### DIFF
--- a/misc/language-tour-guide/src/components/ExampleLayout.tsx
+++ b/misc/language-tour-guide/src/components/ExampleLayout.tsx
@@ -16,7 +16,6 @@ export function ExampleLayout() {
     setIsLargeScreen(mediaQuery.matches);
 
     const handleChange = (e: MediaQueryListEvent) => {
-      console.log("e: ", e);
       setIsLargeScreen(e.matches);
     };
 
--- a/misc/language-tour-guide/src/utils/hooks/useOnNavigationChange.ts
+++ b/misc/language-tour-guide/src/utils/hooks/useOnNavigationChange.ts
@@ -1,6 +1,6 @@
 import { useLocation } from "react-router";
 import { useCodeEditorStore } from "../../store/codeEditor";
-import { useEffect } from "react";
+import { useEffect, useRef } from "react";
 import { useExecutionHistoryStore } from "../../store/executionHistory";
 
 export function useOnNavigationChange(callback: () => void) {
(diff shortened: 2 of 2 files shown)


misc/language-tour-guide/src/components/ExampleLayout.tsx
```diff
@@ -2,15 +2,66 @@ import { CodeEditor } from "./code-editor/CodeEditor";
...
+
+  useEffect(() => {
+    // set media query with 1024px breakpoint which matches Tailwind's lg breakpoint
+    const mediaQuery = window.matchMedia("(min-width: 1024px)");
+
+    // set initial value
+    setIsLargeScreen(mediaQuery.matches);
+
+    const handleChange = (e: MediaQueryListEvent) => {
+      console.log("e: ", e);
```
jgonet: Remove `console.log`.

misc/language-tour-guide/src/utils/hooks/useOnNavigationChange.ts
```diff
@@ -1,15 +1,23 @@
...
+  const clearHistory = useExecutionHistoryStore((state) => state.clearHistory);
 
   useEffect(() => {
     resetStdoutResult();
+    resetStderrResult();
+    clearHistory();
+    window.scrollTo({ top: 0, left: 0, behavior: "instant" });
     callback();
-  }, [pathname, callback, resetStdoutResult]);
+  }, [pathname, callback, resetStdoutResult, resetStderrResult, clearHistory]);
```
jgonet: I'm afraid of glitching if callback isn't memoized.

mmietla: yeah, it will re-render component in the loop

mmietla: I removed `callback` from useEffect deps array

jgonet: This isn't the correct solution – we would use old ref if it ever changed. We can either customize useCallback return type to brand it or use `useRef` to prevent `useEffect` from firing on change but still use newest ref for callback.

PR #375, opened 2025-10-28
Refactor loading the avm bundle in WASM

mat-hek:
Requires https://github.com/software-mansion-labs/FissionVM/pull/160

### DIFF
--- a/priv/static-template/wasm/popcorn_iframe.js
+++ b/priv/static-template/wasm/popcorn_iframe.js
@@ -48,14 +48,14 @@ export async function runIFrame() {
   send(MESSAGES.START_VM, initProcess);
 }
 
-async function startVm(avm_bundle) {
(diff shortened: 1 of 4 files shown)


priv/static-template/wasm/popcorn_iframe.js
```diff
@@ -27,11 +28,37 @@ class TrackedValue {
...
+      console.error(`Iframe: received unhandled popcorn event: ${JSON.stringify(data, null, 4)}`);
+    }
+  });
+
+  send(MESSAGES.START_VM, initProcess);
+}
+
+async function startVm(avm_bundle) {
+  let resolve_result_promise = null;
+  const result_promise = new Promise((resolve) => { resolve_result_promise = resolve; });
```
jgonet: JS uses camelCase.

PR #379, opened 2025-11-03
Langtour: Refactor PopcornProvider and extract evaluation logic

mmietla:
**PopcornProvider Refactoring:**
  - Simplified initialization logic with centralized error handling
  - Introduced `LogSink` interface for outputs (stdout/stderr/crash)

  **Extracted Evaluation Logic (`usePopcornEval`):**
  - Created dedicated hook for code evaluation that encapsulates:
    - Log capture lifecycle (start/stop)
    - Timeout and long-running detection
    - Error handling and Sentry reporting
    - History entry creation

  **Enhanced Logging:**
  - Added log listener registration logic in `popcorn.js`
  - Enables targeted log capture per evaluation instead of global collection

  **Store Cleanup:**
  - Removed stdout/stderr from global `codeEditor` store
  - Moved to local component state where appropriate

### DIFF
--- a/misc/language-tour-guide/src/components/execution/Results.tsx
+++ b/misc/language-tour-guide/src/components/execution/Results.tsx
@@ -30,10 +30,6 @@ export function Results() {
     }))
   );
 
-  const handleCancelRunning = () => {
-    cancelCall();
-  };
-
   const addHistoryEntry = useExecutionHistoryStore(
     (state) => state.addHistoryEntry
   );
@@ -116,11 +112,7 @@ export function Results() {
           )}
         </div>
         {pending && longRunning && (
-          <Button
-            title="Cancel"
-            type="secondary"
-            onClick={handleCancelRunning}
-          />
+          <Button title="Cancel" type="secondary" onClick={cancelCall} />
         )}
         {isCodeChanged() && (
           <Button
--- a/priv/static-template/wasm/popcorn.js
+++ b/priv/static-template/wasm/popcorn.js
@@ -58,7 +58,7 @@ export class Popcorn {
   _requestId = 0;
   _calls = new Map();
   _listenerRef = null;
-  _logListeners = new Set();
+  _logListeners = { stdout: new Set(), stderr: new Set() };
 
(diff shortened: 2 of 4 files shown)


misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -1,159 +1,81 @@
...
 
   const handleCancelRunning = () => {
-    try {
-      reinitializePopcorn();
-      console.log("Cancel running code");
-    } catch (e) {
-      console.error("Error during Popcorn reinitialization:", e);
-    }
+    cancelCall();
   };
```
jgonet: We can use `cancelCall` directly.

priv/static-template/wasm/popcorn.js
```diff
@@ -240,12 +241,39 @@ export class Popcorn {
...
+    this._logListeners.delete(listener);
+  }
+
+  _notifyLogListeners(output) {
+    this._logListeners.forEach((listener) => {
+      try {
+        listener(output);
+      } catch (error) {
+        console.error("Error in log listener:", error);
+      }
```
jgonet: I feel like we should drop try-catch here.

priv/static-template/wasm/popcorn.js
```diff
@@ -240,12 +241,39 @@ export class Popcorn {
...
+   * Registers a log listener that will be called when stdout or stderr output is received.
+   * @param {function({stdout?: string, stderr?: string}): void} listener - Callback function
+   */
+  registerLogListener(listener) {
+    this._logListeners.add(listener);
+  }
+
+  /**
+   * Unregisters a previously registered log listener.
+   * @param {function({stdout?: string, stderr?: string}): void} listener - Callback function to remove
```
jgonet: I don't like conflating stdout with stderr, maybe `registerLogListener(callback, "stdout" | "stderr")`? This way we don't need to do checks in React for `output.stdout !== undefined`.

jgonet (approved):

PR #385, opened 2025-11-04
Langtour: Add a loader when Popcorn is initializing

mmietla:
Add a loader while Popcorn is initializing and the bundle is not fully loaded.

https://github.com/user-attachments/assets/32e191a7-4f35-40c5-94c9-dd37dd74801a

### DIFF
--- a/misc/language-tour-guide/src/components/execution/PopcornLoader.tsx
+++ b/misc/language-tour-guide/src/components/execution/PopcornLoader.tsx
@@ -1,0 +1,22 @@
+import { usePopcorn } from "../../context/popcorn/actions";
+
+import Logo from "../../assets/logo.svg?react";
+
+export function PopcornLoader() {
+  const { isLoadingPopcorn } = usePopcorn();
+
+  return (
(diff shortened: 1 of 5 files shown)


misc/language-tour-guide/src/components/execution/PopcornLoader.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <div className="bg-light-30/90 absolute inset-0 z-10 flex items-center justify-center rounded-md backdrop-blur-sm">
+          <div className="flex flex-col items-center gap-4">
+            <Logo className="h-16 w-16 animate-bounce" />
+            <p className="text-dark-50 font-mono text-sm font-medium">
+              Initializing Popcorn...
+            </p>
+          </div>
+        </div>
+      )}
+    </>
```
jgonet, line 20: It should probably be `if(!isLoadingPopcorn) return null`.
The better solution would be to avoid absolute containers and render Results and History conditionally.

PR #395, opened 2025-11-07
Patch erlang:fun_info_mfa

mat-hek:
This enables `Task.start_link`

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,5 +1,4 @@
 # Used by "mix format"
 [
-  inputs: ["{mix,.formatter}.exs", "{config,lib,test,patches}/**/*.{ex,exs}"],
-  subdirectories: ["examples/*"]
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test,patches}/**/*.{ex,exs}"]
 ]
--- a/patches/otp/erts/erlang.erl
+++ b/patches/otp/erts/erlang.erl
@@ -111,6 +111,7 @@
     nif_error/1,
     bump_reductions/1,
     fun_info/1,
+    fun_info_mfa/1,
     phash2/1,
     phash2/2,
     phash/2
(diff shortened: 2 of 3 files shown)


patches/otp/erts/erlang.erl
```diff
@@ -3176,6 +3177,9 @@ fun_info(Fun) ->
     Info = [?MODULE:fun_info(Fun, Key) || Key <- [module, name, arity, type, env]],
     Info ++ [{new_index, 0}, {uniq,  erlang:unique_integer()}].
 
+fun_info_mfa(Fun) ->
+    {?MODULE:fun_info(Fun, module), ?MODULE:fun_info(Fun, name), ?MODULE:fun_info(Fun, arity)}.
```
jgonet, line 3181: Undocumented and implemented as NIF in OTP? Fun.

.formatter.exs
```diff
@@ -1,5 +1,4 @@
 # Used by "mix format"
 [
-  inputs: ["{mix,.formatter}.exs", "{config,lib,test,patches}/**/*.{ex,exs}"],
-  subdirectories: ["examples/*"]
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test,patches}/**/*.{ex,exs}"]
```
jgonet, line 3: Why this change?

mat-hek: Otherwise mix format also formats examples, and the liveview example has phoenix in the formatter deps and formatting fails since phoenix is not a dependency of popcorn

patches/otp/erts/erlang.erl
mat-hek: Have a look at 'erts_internal.erl'...

.formatter.exs
mat-hek: Here's what happens without this change: https://github.com/software-mansion/popcorn/actions/runs/19166714172/job/54789030104

PR #401, opened 2025-11-13
Langtour: sentry session health

mmietla:
- Add integration for collecting session health.
- Change captureMessage to captureException to properly count errors in Sentry session health.


jgonet (approved):
💯

PR #407, opened 2025-11-17
Fixed demos on landing page deploy

FranekKubis:
BundlePaths for .avm bundles updated.
Verified on locally run docker container. (base.Dockerfile -> landing.Dockerfile -> localhost)

### DIFF
--- a/misc/landing-page/src/pages/demos/eval.astro
+++ b/misc/landing-page/src/pages/demos/eval.astro
@@ -135,7 +135,7 @@ end"
 
     try {
       popcorn = await window.Popcorn.init({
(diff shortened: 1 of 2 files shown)


FKubisSWM:
> Why it was failing with .beam parse error? Can we create some checks to make it easier when you don't pass right file?

I agree we should check if we can access bundle file while we are creating popcorn iframe.

What do you think.about using fetch function for checking the path while initiating popcorn iframe?

example:

```
fetch(this._bundlePath, { method: 'HEAD' })
.then(response => {
    if (response.ok) {
      //everythings fine
    } else {
      //handle missing resource
    }
  })
```

Issue #415, opened 2025-11-18
Allow using .livemd files


jgonet:
`.livemd` format is specific to Livebook, stores cells and optional execution result. We should be able to use it to write tutorial pages.

Implementation should load `.livemd` and transform it to `.mdx` where we will mount editors. Editor changes are covered in different issues (#383, #409). Here, we should use dummy component. We should also research if we can integrate this components directly with current code or will it need some changes (only creating component that have id and renders editor itself based on available contexts, etc).

PR #416, opened 2025-11-18
Langtour: Add livemd support with parser and Vite plugin

mmietla:
Implements: #415

### DIFF
--- a/misc/language-tour-guide/src/plugins/livemd/index.ts
+++ b/misc/language-tour-guide/src/plugins/livemd/index.ts
@@ -24,13 +24,12 @@ export function livemdPlugin(): Plugin {
         return;
       }
 
-      const livemdName = id.replace(/\.mdx$/, "");
-      const livemdPath = path.join(rootPath, "src", livemdName);
+      const livemdName = path.basename(id).replace(/\.mdx$/, "");
+      const livemdPath = path.join(rootPath, "src", "content", livemdName);
 
-      // Read the .livemd file
       const livemdContent = await readFile(livemdPath, "utf-8");
 
-      return String(await transformToMdx(livemdContent));
+      return transformToMdx(livemdContent);
     }
   };
 }
--- a/misc/language-tour-guide/src/plugins/livemd/parser.ts
+++ b/misc/language-tour-guide/src/plugins/livemd/parser.ts
@@ -135,5 +135,5 @@ export async function transformToMdx(markdown: string) {
       fences: true
     });
 
-  return await processor.process(markdown);
+  return String(await processor.process(markdown));
 }


misc/language-tour-guide/src/plugins/livemd/index.ts
```diff
@@ -0,0 +1,38 @@
...
+        return source + ".mdx";
+      }
+    },
+
+    async load(id) {
+      if (id.endsWith(".livemd.mdx")) {
+        const livemdPath = id.replace(/\.mdx$/, "");
+
+        // Read the .livemd file
+        const livemdContent = fs.readFileSync(
```
jgonet: We should do async reading, no?

misc/language-tour-guide/src/plugins/livemd/parser.ts
```diff
@@ -0,0 +1,142 @@
...
+          parent.children.splice(index, 1);
+        }
+      }
+    });
+  };
+}
+
+function escapeBackticks(value: string): string {
+  return value.replace(/`/g, "\\`");
+}
```
jgonet: This could be a variable.

misc/language-tour-guide/src/plugins/livemd/index.ts
```diff
@@ -0,0 +1,38 @@
...
+          path.join(rootPath, "src", livemdPath),
+          "utf-8"
+        );
+
+        const result = await parseLivemd(livemdContent);
+
+        return String(result);
```
jgonet: 1. We can simplify it if we return string from `parseLivemd`.
2. nit: maybe rename to `transformToMdx`/`asMdxSource`/etc?

misc/language-tour-guide/src/plugins/livemd/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+
+    resolveId(source) {
+      if (source.endsWith(".livemd")) {
+        return source + ".mdx";
+      }
+    },
+
+    async load(id) {
+      if (id.endsWith(".livemd.mdx")) {
```
jgonet: Nit: early return to drop an indent.

misc/language-tour-guide/src/plugins/livemd/index.ts
```diff
@@ -0,0 +1,38 @@
...
+        const livemdPath = id.replace(/\.mdx$/, "");
```
jgonet: nit: maybe:
```js
const livemdName = id.replace(/\.mdx$/, "");
const livemdPath = path.join(rootPath, "src", livemdPath);

const livemd = asyncRead(livemdPath, "utf-8");
return parseLivemd(livemd);
```

misc/language-tour-guide/src/plugins/livemd/index.ts
mmietla: should be, I missed it

misc/language-tour-guide/src/plugins/livemd/parser.ts
mmietla: You mean something like this?

```
const escapeBackticks = (v: string) => v.replace(/`/g, "\\`");
```

jgonet:
```js
function wrapProps(value: string): string {
  const escaped = value.replace(/`/g, "\\`");
  return "`" + escaped + "`";
}
```

jgonet: Up to you, though.

misc/language-tour-guide/src/plugins/livemd/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    async load(id) {
+      if (!id.endsWith(".livemd.mdx")) {
+        return;
+      }
+
+      const livemdName = id.replace(/\.mdx$/, "");
+      const livemdPath = path.join(rootPath, "src", livemdName);
+
+      // Read the .livemd file
```
jgonet:
```suggestion
```

misc/language-tour-guide/src/plugins/livemd/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+  const processor = unified()
+    .use(remarkParse)
+    .use(remarkLivebook)
+    .use(remarkStringify, {
+      bullet: "-",
+      fence: "`",
+      fences: true
+    });
+
+  return await processor.process(markdown);
```
jgonet: This await doesn't do anything except for wrapping in another promise. Maybe you wanted `return String(await ...)`?

jgonet (approved):

PR #418, opened 2025-11-20
Patch binary:decode_unsigned, binary:encode_unsigned and fix printing floats

mat-hek:


misc/language-tour-guide/elixir_tour/config/.gitignore
```diff
@@ -0,0 +1 @@
+*.secret.exs
```
jgonet: We can ignore it in the main file.

PR #420, opened 2025-11-25
LocalLiveView Forms Demo

FranekKubis:
From README:

> The demo form is build as a html page with a single form and a rendered list of users (usernames and emails). 
List of users is stored in local live view process memory.
The form handles onchange and onsubmit LocalLiveView events.
The change event triggers validate function that looks through already saved users and makes sure that
new username or email isn't already present in the users list. The submit event add a new user to the user list.


examples/form_demo/README.md
```diff
@@ -0,0 +1,35 @@
+# FormDemo
+
+The demo project implementing phoenix forms using LocalLiveView.
+
+The demo form is build as a html page with a single form and a rendered list of users (usernames and emails). 
+List of users is stored in local live view process memory.
+The form handles onchange and onsubmit LocalLiveView events.
+The change event triggers validate function that looks through already saved users and makes sure that
+new username or email isn't already present in the users list. The submit event add a new user to the user list.
```
mat-hek:
```suggestion
new username or email isn't already present in the users list. The submit event adds a new user to the user list.
```

examples/form_demo/README.md
```diff
@@ -0,0 +1,35 @@
+# FormDemo
+
+The demo project implementing phoenix forms using LocalLiveView.
+
+The demo form is build as a html page with a single form and a rendered list of users (usernames and emails).
```
mat-hek:
```suggestion
The demo form is built as an html page with a single form and a rendered list of users (usernames and emails). 
```

examples/form_demo/README.md
```diff
@@ -0,0 +1,35 @@
+# FormDemo
```
mat-hek:
```suggestion
# Local Live View Form Demo
```

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -0,0 +1,67 @@
+defmodule FormDemoLocal do
+  use LocalLiveView
+  import Local.CoreComponents
+
+  def render(assigns) do
```
mat-hek, line 6:
```suggestion
  @impl true
  def render(assigns) do
```

same for other callbacks

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -0,0 +1,67 @@
...
+      user
+      |> Enum.reduce(%{}, fn {k, v}, acc ->
+        in_existing_users = Enum.any?(existing_users, fn u -> Map.get(u, k) == v end)
+        Map.put(acc, k, in_existing_users)
+      end)
+
+    Enum.reduce(result, "", fn
+      {k, true}, acc -> acc <> "#{to_string(k)} already in use! "
+      {_k, false}, acc -> acc
+    end)
```
mat-hek:
```suggestion
      |> Enum.filter(fn {key, value} ->
        Enum.any?(existing_users, fn user -> Map.get(user, key) == value end)
      end)
      |> Enum.map_join(", ", fn {key, _value} -> "#{key} already in use") end)
      |> String.capitalize()
```

examples/local_live_view/priv/static/local_live_view.js
```diff
@@ -1,30 +1,55 @@
 import { Popcorn } from "./wasm/popcorn.js";
 
+var events = [
+    {js_event: "click", pop_event: "pop-click"},
+    {js_event: "keydown", pop_event: "pop-keydown"},
+    {js_event: "submit", pop_event: "pop-submit"},
+    {js_event: "input", pop_event: "pop-change"}
```
mat-hek, line 7: How hard would it be to support more events?

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -0,0 +1,67 @@
...
+
+  def render(assigns) do
+    ~H"""
+    <.form for={@form} id="my-form" pop-change="validate" pop-submit="save" class="bordered">
+      <label>USERNAME</label>
+      <.input type="text" field={@form[:username]} />
+      <label>EMAIL</label>
+      <.input type="email" field={@form[:email]} />
+      <div class="centered">
+        <button class="ghost-button">SAVE</button>
```
mat-hek: Can we easily disable the button when validation doesn't pass?

examples/local_live_view/priv/static/local_live_view.js
FranekKubis: It should be as easy as adding an event to this list ...

... but:
It will not work exactly the same as Phoenix Live View.
Some events (like input for example) need special handling in javascript part of the Phoenix Live View.

How it is done now:
The local_live_view.js adds event listener on the js_event on the element where pop_event attribute is present. The event listener handler sends event to LocalLiveView via popcorn.call.

How I think it should be done:
The best way would be to use modified phoenix_live_view.js instead of local_live_view.js to handle the events the same way phoenix does it. 
Plan was to modify phoenix_live_view.js to instead of sending events via liveSocket to server send them to AtomVM via popcorn.call. I did not do this because it would take much more time and effort than adding simple event listener - so I left it "for the future".

TL DR:
It should be easy but some events could need additional special treatment.

mat-hek: Got it, please create an issue for that

FranekKubis: https://github.com/software-mansion/popcorn/issues/435

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -52,16 +55,12 @@ defmodule FormDemoLocal do
...
-    Enum.reduce(result, "", fn
-      {k, true}, acc -> acc <> "#{to_string(k)} already in use! "
-      {_k, false}, acc -> acc
+    user
+    |> Enum.filter(fn {key, value} ->
+      Enum.any?(existing_users, fn user -> Map.get(user, key) == value end)
     end)
+    |> Enum.map_join(", ", fn {key, _value} ->
+      String.capitalize("#{key} already in use")
+      end)
```
mat-hek: Seems like a formatting issue. Please add `mix format --check-formatted` to CI

mat-hek (changes requested):
- fix email input error
- disallow adding an empty user
- fix warnings when compiling the example

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -33,34 +35,74 @@ defmodule FormDemoLocal do
...
+  defp validate_correctness("email", value) do
+    case String.split(value, "@") do
+      [name, server] ->
+        case String.length(name) > 0 and String.contains?(server, ".") do
+          true -> ""
+          false -> "Email must have an email format"
+        end
+
+      _ ->
+        "Email must have an email format"
```
mat-hek: let's use `with`

mat-hek (approved):

Issue #383, opened 2025-11-04
Single page layout


jgonet:
Currently, Elixir language tour looks like that:

<details open><summary><b>Screenshot – current</b></summary>
<p>

<img width="1319" height="1177" alt="Image" src="https://github.com/user-attachments/assets/501551f9-6df9-4269-a251-c8e75b2809e3" />

</p>
</details> 

There's few issues discovered with that:
- referencing code in text is harder, person reading needs to look for the comments in the code.
- isolating snippets is impossible, we execute everything as one code block.
- we don't have an ability to create incomplete snippets (this is attractive didactically but may be extracted to other issue).

We'd like to create singe-column layout, inspired by livebook.

This is a variant generated by Claude:
<details open><summary><b>Screenshot – variant one</b></summary>
<p>

<img width="935" height="877" alt="Image" src="https://github.com/user-attachments/assets/cd2d83fa-f83a-4627-a4e3-0bd086188637" />

</p>
</details>

Aside of changing CSS, we need to allow multiple editors and interactions between them. This is tracked by #409.

PR #426, opened 2025-11-28
Langtour: single page layout

mmietla:
Implements: #383 

image
image

### DIFF
--- a/misc/language-tour-guide/src/components/execution/Results.tsx
+++ b/misc/language-tour-guide/src/components/execution/Results.tsx
@@ -1,9 +1,9 @@
-import { useState, useEffect } from "react";
 import { useEditorResult, useEditorExecuting } from "../../store/editors";
 import StdoutResults from "../execution/StdoutResults";
 import { CompilerError } from "../execution/CompilerError";
 import { WarningOutput } from "../execution/WarningOutput";
 import ArrowRight from "../../assets/arrow-right.svg?react";
+import { useDelayedPending } from "../../utils/hooks/useDelayedPending";
 
 type EditorResultsProps = {
   id: string;
@@ -12,18 +12,7 @@ type EditorResultsProps = {
 export function Results({ id }: EditorResultsProps) {
   const result = useEditorResult(id);
   const isExecuting = useEditorExecuting(id);
-  const [longExecuting, setLongExecuting] = useState(false);
-
-  useEffect(() => {
-    if (isExecuting) {
-      const timer = setTimeout(() => {
-        setLongExecuting(true);
-      }, 300);
-      return () => clearTimeout(timer);
-    } else {
-      setLongExecuting(false);
-    }
-  }, [isExecuting]);
+  const longExecuting = useDelayedPending(isExecuting);
 
   if (!result) {
     return null;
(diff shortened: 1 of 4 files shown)


misc/language-tour-guide/src/store/editors.ts
```diff
@@ -0,0 +1,143 @@
...
+        }
+      });
+    },
+
+    setEditorResult: (id: string, result) => {
+      set((state) => {
+        const editor = state.editors.get(id);
+        if (editor) {
+          editor.currentResult = {
+            ...result
```
jgonet: Maybe `structuredClone`?

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -1,146 +1,79 @@
...
+type EditorResultsProps = {
+  id: string;
+};
+
+export function Results({ id }: EditorResultsProps) {
+  const result = useEditorResult(id);
+  const isExecuting = useEditorExecuting(id);
+  const [longExecuting, setLongExecuting] = useState(false);
 
   useEffect(() => {
```
jgonet: This is kinda convoluted. What do you think about something like that?

```ts
useDebouncedEffect(isExecuting, (isExecutingStable) => {
  if (isExecutingStable) setLongExecuting(true);
}, {ms: 300})

useEffect(() => {
  if (!isExecuting) setLongExecuting(false);
}, [isExecuting]);
```

Not sure if `useDebouncedEffect` could be used in other places with setTimeouts as well, maybe it's too overengineered.

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -1,146 +1,79 @@
...
-    };
-  }, [onKeyDown]);
+  return (
+    <div className="text-brown-90 bg-light-30 border-grey-20 flex flex-col gap-2 rounded-b-md border-t px-4 py-3">
+      <div className="text-brown-90/60 flex items-center gap-2 text-sm">
+        <span className="">Output</span>
+        <span className="text-xs font-light">
+          {longExecuting
+            ? "pending..."
+            : durationMs !== undefined && `${durationMs.toFixed(3)}ms`}
```
jgonet, line 42: Capitalize pending.

misc/language-tour-guide/src/components/markdown/MdxWrapper.tsx
```diff
@@ -1,35 +1,31 @@
...
-  }, [
-    code,
-    setCode,
-    setDefaultCode,
-    getCodeFromStorage,
-    setPathHash,
-    pathname
-  ]);
+    for (const snippet of codeSnippets) {
+      initEditor(snippet.id, snippet.initCode, snippet.output, snippet.stdout);
```
jgonet: `initEditor` could take an object.

misc/language-tour-guide/src/components/execution/Results.tsx
mmietla: We have similar logic in executing code where the cancel button is displayed after 500ms. We can remove this timeout from evalCode, and manage it in the component with this DebouncedEffect custom hook.

Issue #409, opened 2025-11-18
Multiple editors


jgonet:
In undergoing refactor of langtour we need an ability to create multiple editors, livebook-style.

We need them to be able to:
0. share state and variables
1. mark all following editors as stale if code changes
2. track execution of previous editors, running them if necessary.

## Stale editors

When changing code in editor, there's a possibility of invalidating next editors in the page that may use variables defined in the current one. Detecting whether they need to be invalidated is quite tricky, since we'd need to track execution environment (process state, env state, processes) and variables.

We should start by invalidating all editors in the page following edited one.

## Tracking execution

Editor can be in four states:
- not yet run
- run (success)
- run (failure)
- stale

When running code in the editor we need to look at all previous ones. We should run all editors from first that didn't succeed or it's stale to the current one.

Examples [^1]:
`[✓, ✓, _, [_], _]` – when running fourth editor, we should run third beforehand and leave fifth in "not yet run" state.
`[✓, X, _, _, [_]]` – when running fifth editor, we should start from second since it failed.
`[✓, ✓, _, S, [S]]` – after editing in third editor, two following are stale. If we run last one, we should run third and fourth beforehand.

[^1]: `[ ]` – currently run cell,
`_` – not yet run,
`✓` – run (success),
`X` – run (failure),
`S` – stale

PR #430, opened 2025-12-02
Langtour: multiple editors

mmietla:
Implements: #409

https://github.com/user-attachments/assets/21a07167-1c5c-4cf8-852b-25868c5bde36

### DIFF
--- a/misc/language-tour-guide/src/components/code-cell/CodeDisplay.tsx
+++ b/misc/language-tour-guide/src/components/code-cell/CodeDisplay.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useRef, useEffect } from "react";
+import { useCallback, useRef, useEffect, useState } from "react";
 import { Button } from "../Button";
 import { CodeEditor } from "./CodeEditor";
 import {
@@ -26,19 +26,18 @@ export default function CodeDisplay({ id }: CodeDisplayProps) {
   const isCodeChanged = useEditorChanged(id);
   const longRunning = useDelayedPending(isExecuting || isQueued, 300);
   const executionState = useEditorExecutionState(id);
-
-  const lastStableStateRef = useRef(executionState);
+  const [lastStableState, setLastStableState] = useState(executionState);
 
   useEffect(() => {
     if (executionState !== "running" && executionState !== "queued") {
-      lastStableStateRef.current = executionState;
+      setLastStableState(executionState);
     }
   }, [executionState]);
 
   const displayState =
     (executionState === "running" || executionState === "queued") &&
     !longRunning
-      ? lastStableStateRef.current
+      ? lastStableState
       : executionState;
 
   const {
--- a/misc/language-tour-guide/src/store/editors.ts
+++ b/misc/language-tour-guide/src/store/editors.ts
@@ -145,9 +145,6 @@ export const useEditorsStore = create<EditorsStore>()(
     getEditorsToRun: (id: string) => {
       const state = get();
       const currentIndex = state.editorOrder.indexOf(id);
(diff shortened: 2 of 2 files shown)


misc/language-tour-guide/src/components/code-cell/CodeDisplay.tsx
```diff
@@ -1,33 +1,60 @@
...
-  const setEditorResult = useEditorsStore((state) => state.setEditorResult);
-  const resetEditorToDefault = useEditorsStore(
-    (state) => state.resetEditorToDefault
+  const lastStableStateRef = useRef(executionState);
+
+  useEffect(() => {
+    if (executionState !== "running" && executionState !== "queued") {
+      lastStableStateRef.current = executionState;
+    }
+  }, [executionState]);
```
jgonet, line 35: Why do we need ref for that?

misc/language-tour-guide/src/components/execution/Results.tsx
```diff
@@ -33,13 +38,15 @@ export function Results({ id }: EditorResultsProps) {
...
+      className={`text-brown-90 bg-light-30 border-grey-20 flex flex-col gap-2 rounded-b-md border-t px-4 py-3 ${isQueued ? "opacity-60" : ""}`}
+    >
       <div className="text-brown-90/60 flex items-center gap-2 text-sm">
         <span className="">Output</span>
         <span className="text-xs font-light">
-          {longExecuting
-            ? "Pending..."
-            : durationMs !== undefined && `${durationMs.toFixed(3)}ms`}
+          {!longExecuting &&
+            durationMs !== undefined &&
```
jgonet, line 48: When `durationMs` is undefined?

misc/language-tour-guide/src/store/editors.ts
```diff
@@ -92,23 +117,67 @@ export const useEditorsStore = create<EditorsStore>()(
...
+
+        if (editor && editor.executionState !== "success") {
+          startIndex = i;
+          break;
+        }
+      }
+
+      for (let i = startIndex; i <= currentIndex; i++) {
+        editorsToRun.push(state.editorOrder[i]);
+      }
```
jgonet: We could slice `state.editorOrder`.

misc/language-tour-guide/src/store/editors.ts
```diff
@@ -92,23 +117,67 @@ export const useEditorsStore = create<EditorsStore>()(
...
+            editor.executionState = "stale";
+          }
+        }
+      });
+    },
+
+    getEditorsToRun: (id: string) => {
+      const state = get();
+      const currentIndex = state.editorOrder.indexOf(id);
+      if (currentIndex === -1) return [id];
```
jgonet: When this happens? I was under impression `state.editorOrder` always contain all editors' ids.

misc/language-tour-guide/src/components/code-cell/CodeDisplay.tsx
mmietla: Yeah, it is unnecessary. We can replace it with useState

misc/language-tour-guide/src/components/execution/Results.tsx
mmietla: When there is saved output in livemd, in this situation we have only output and stdoutResults

misc/language-tour-guide/src/store/editors.ts
mmietla: I missed that we can omit this condition because when there is a wrong editor ID, we will simply return an empty array, so the wrong editor doesn't run.

Issue #424, opened 2025-11-27
Add configuration file for content ordering


mmietla:
Currently, content ordering in the language tour guide is done by multiple mechanisms:
1. Directory naming with numeric prefixes
2. Frontmatter with order in individual files

### Solution
Centralized configuration file that defines:
1. The order and structure of all content
2. Metadata for each content piece
3. Navigation tree structure

PR #436, opened 2025-12-05
Langtour: configuration file

mmietla:
implements: #424

### DIFF
--- a/misc/language-tour-guide/src/components/navigation/NavigationBar.tsx
+++ b/misc/language-tour-guide/src/components/navigation/NavigationBar.tsx
@@ -1,7 +1,5 @@
 import { NavigationButton } from "./NavigationButton";
-// import { useSiblingsNode } from "../../utils/hooks/useSiblingsNode";
 import { FeedbackButton } from "./FeedbackButton";
-// import { getNodeNavigationSiblings } from "../../utils/content/navigation-builder";
 import { useLocation } from "react-router";
 import { getNodeNavigationSiblings } from "../../utils/content/navigation-builder";
 


misc/language-tour-guide/src/utils/content/configuration.ts
```diff
@@ -0,0 +1,54 @@
...
+    };
+  });
+
+  return LazyComponent;
+};
+
+export const navigationConfig: NavigationConfig = [
+  {
+    group: "Basic types and operations",
+    slug: "basic-types-and-operations",
```
jgonet: Maybe we could calculate slug based on name?

misc/language-tour-guide/src/utils/content/navigation-builder.ts
```diff
@@ -1,119 +1,113 @@
...
+  indexInParent: number;
+};
 
-function compareEntriesByOrder(a: [Dir, DirEntry], b: [Dir, DirEntry]): number {
-  const [, entryA] = a;
-  const [, entryB] = b;
+function flattenTreeToLinks(
+  tree: NavigationTreeItem[]
+): FlattenedNavigationItem[] {
+  console.log(tree);
```
jgonet: Console.log

misc/language-tour-guide/src/utils/content/types.ts
```diff
@@ -1,40 +1,57 @@
...
+export function isConfigItem(
+  item: ConfigItem | ConfigGroup
+): item is ConfigItem {
+  return (item as ConfigItem).component !== undefined;
+}
+
+export function isConfigGroup(
+  item: ConfigItem | ConfigGroup
+): item is ConfigGroup {
+  return (item as ConfigGroup).items !== undefined;
```
jgonet, line 14: Or `hasOwnProperty`, `hasOwn`.

misc/language-tour-guide/src/components/markdown/MdxWrapper.tsx
```diff
@@ -1,15 +1,17 @@
...
-type MdxWrapperProps = {
+export type MdxWrapperProps = {
   Component: React.ComponentType;
   codeSnippets?: CodeSnippet[];
 };
 
 export function MdxWrapper({ Component, codeSnippets }: MdxWrapperProps) {
   const initEditor = useEditorsStore((state) => state.initEditor);
   const [isInitialized, setIsInitialized] = useState(false);
+  useOnNavigationChange();
```
jgonet: This is kinda weird, why we need that?

mmietla: I thought it would be nice to group the navigation change logic into a single hook, but we can also unwrap it and use it directly in this component

misc/language-tour-guide/src/utils/content/configuration.ts
mmietla: I agree. I was also thinking about creating component property automatically based on the path to `.livemd` file, so the configuration file would only contain the navigation description without any function invocations

jgonet: Both works for me.

misc/language-tour-guide/src/components/markdown/MdxWrapper.tsx
jgonet: I'm a bit opposed to adding opaque effect hooks into otherwise pure components – it's harder to track when something happens.

misc/language-tour-guide/src/components/navigation/NavigationBar.tsx
```diff
@@ -1,9 +1,15 @@
 import { NavigationButton } from "./NavigationButton";
-import { useSiblingsNode } from "../../utils/hooks/useSiblingsNode";
+// import { useSiblingsNode } from "../../utils/hooks/useSiblingsNode";
 import { FeedbackButton } from "./FeedbackButton";
+// import { getNodeNavigationSiblings } from "../../utils/content/navigation-builder";
```
jgonet: Do we need those comments?

mmietla: nope :sob:

PR #439, opened 2025-12-09
Langtour editors saving

mmietla:
Implements localStorage-based persistence for code editors:

- **Editor state persistence**: Saves modified editor code and execution state to localStorage with debouncing (500ms)
- **Automatic restoration**: Restores saved editor state on page reload or navigation
- **Content hash validation**: Clears localStorage when content changes to prevent stale data

### DIFF
--- a/misc/language-tour-guide/src/vite-env.d.ts
+++ b/misc/language-tour-guide/src/vite-env.d.ts
@@ -1,6 +1,11 @@
 /// <reference types="vite/client" />
 /// <reference types="vite-plugin-svgr/client" />
 
+/**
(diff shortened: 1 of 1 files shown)


misc/language-tour-guide/src/vite-env.d.ts
```diff
@@ -1,6 +1,8 @@
 /// <reference types="vite/client" />
 /// <reference types="vite-plugin-svgr/client" />
 
+declare const __CONTENT_HASH__: string;
```
jgonet, line 9: Would be nice to document what it does.

PR #440, opened 2025-12-10
Local live view phoenix connection

FranekKubis:
From README.md:

> The demo project implementing phoenix forms using LocalLiveView.
> 
> The demo form is built as an html page with a single form and a two rendered list of users (usernames and emails).
> 
> One list is stored in the local LiveView assigns, while the other is stored in the Phoenix LiveView assigns. These will be referred to as the local list and the server list, respectively.
> 
> The form handles onchange and onsubmit LocalLiveView events. The change event triggers validate function that looks through already saved users and makes sure that new username or email isn't already present in the users list. The submit event adds a new user to the user list.
> 
> On mount both local and phoenix liveViews synchronizes their states (list of users).
> 
> On submit new user is added to the local list and also sent to server list.
> 
> **Since both lists are stored in the runtime, the demo illustrates the persistent nature of a local-first approach. You can shut down the server to see the server list synchronize from its local counterpart upon restart. Conversely, refreshing the page will empty the local list, which then synchronizes from the server.**

### DIFF
--- a/examples/local_live_view/priv/static/local_live_view.js
+++ b/examples/local_live_view/priv/static/local_live_view.js
@@ -19,20 +19,6 @@ async function setup() {
   const { data, durationMs } = await popcorn.call({ "views": find_predefined_views()}, {
     timeoutMs: 10_000,
   });
-  
-  window.addEventListener("phx:llv_server_message", async (e) => {
-    const { data, durationMs } = await popcorn.call({ "view": e.detail.view, "event": "llv_server_message", 
-        "payload": { "event": "llv_server_message", "value": e.detail.payload } }, {
-        timeoutMs: 10_000,
-      });
-  });
-  
-  window.addEventListener("phx:llv_rerender", async (e) => {
-    const { data, durationMs } = await popcorn.call({ "view": e.detail.view, "event": "llv_rerender", 
-        "payload": { "event": "llv_rerender" } }, {
-        timeoutMs: 10_000,
-    });
-  });
 }
 
 async function afterRenderBind(renderEvent, popcorn) {
@@ -49,6 +35,18 @@ async function afterRenderBind(renderEvent, popcorn) {
       }
     });
   })
+  window.addEventListener("phx:llv_server_message", async (e) => {
+    const { data, durationMs } = await popcorn.call({ "view": e.detail.view, "event": "llv_server_message", 
+      "payload": { "event": "llv_server_message", "value": e.detail.payload } }, {
+      timeoutMs: 10_000,
+    });
+  });
+  window.addEventListener("phx:llv_rerender", async (e) => {
+    const { data, durationMs } = await popcorn.call({ "view": e.detail.view, "event": "llv_rerender", 
+      "payload": { "event": "llv_rerender" } }, {
+      timeoutMs: 10_000,
+    });
+  });
   view_element.dataset.popcornBound = "true";
 }
 


mat-hek (commented):
As discussed, let's add a way to test synchronization and add 'add random' button

examples/form_demo/local/mix.exs
```diff
@@ -26,15 +27,15 @@ defmodule Local.MixProject do
...
+      {:phoenix, "~> 1.8.0", only: [:fmt], runtime: false},
+      {:phoenix_live_view, "~> 1.1.0", only: [:fmt], runtime: false}
     ]
   end
 
   defp aliases do
     [
+      lintformat: "format --check-formatted",
       lint: [
-        "format --check-formatted",
```
mat-hek, line 37: why this change?

examples/form_demo/lib/form_demo_web/live/form_demo_live.ex
```diff
@@ -0,0 +1,61 @@
+defmodule FormDemoWeb.FormDemoLive do
+  use Phoenix.LiveView
+
+  def render(assigns) do
+    ~H"""
+    <div class="centered-div">
+      <div data-pop-view="FormDemoLocal" phx-hook="ServerSendHook" id="FormDemoLocal"></div>
```
kraleppa, line 7: Not sure if this is a right moment for API improvements (it you decide that it's not then you can ignore this comment), but I think ideally would be to use local liveviews in heex similarly to LiveComponents, so like this:
```suggestion
      <.local_live_view module={FormDemoLocal} id="form-demo-local" />
```

examples/form_demo/local/mix.exs
FranekKubis: because locel_live_view does phoenixes module redefinition which generates warnings
phoenix and phoenix_live_view are needed in deps only for formatting
thats why i moved format to build env and made sure that phoenix and phoenix_live_view are present only in this env (:fmt)

examples/form_demo/lib/form_demo_web/live/form_demo_live.ex
```diff
@@ -0,0 +1,61 @@
...
+  def handle_info(:sync, socket) do
+    {:noreply, sync_local_users(socket)}
+  end
+
+  defp sync_local_users(socket) do
+    users = Application.get_env(FormDemo, :users, [])
+    payload = %{"type" => "synchronize", "users" => users}
+
+    socket =
+      push_event(socket, "llv_server_message", %{"view" => "FormDemoLocal", "payload" => payload})
```
kraleppa, line 57: I think it'd good to pass the id of given LocalLiveView since I assume that there might be potentially many LocalLiveViews in a single view defined by the same module

So maybe like this?
```suggestion
       push_event(socket, FormDemoLocal, "form-demo-local", payload)
```

Additional API suggestion: it may be a good to somehow wrap this push_event Again we can take a look at LiveComponents that have `send_update/3` function - maybe it'd be good to create function for LocalLiveViews called `send_local_update`?  Or something like this

examples/form_demo/lib/form_demo_web/live/form_demo_live.ex
```diff
@@ -0,0 +1,61 @@
...
+    {:noreply, socket}
+  end
+
+  def handle_info(:sync, socket) do
+    {:noreply, sync_local_users(socket)}
+  end
+
+  defp sync_local_users(socket) do
+    users = Application.get_env(FormDemo, :users, [])
+    payload = %{"type" => "synchronize", "users" => users}
```
kraleppa, line 54: I assume that you send this synchronize event just after mount and render to set the initial state of LocalLiveView, right? More elegant way is to pass the inital state via attributes, so in the heex you could have

```
<.local_live_view module={FormDemoLocal} id="form-demo-local" users={@users} />
```

And now users could be passed to LocalLiveView via `session` argument in `mount` callback like in nested LiveViews (here is example from [LiveDebugger](https://github.com/software-mansion/live-debugger/blob/c5cac698549b2b27c2fcfb775e53267f770e9c17/lib/live_debugger/app/debugger/async_jobs/web/async_jobs_live.ex#L54)

It's worth considering how it should behave when `@users` assign changed. We could theoretically go with LiveComponent approach and add `update` callback in LocalLiveViews that will be triggered when attributes of given LocalLiveView changed. But you can also go with approach similar to nested liveviews, that does not have `update` callback and when such assign is changed you simply have to send message (usually via pubsub) to the nested LiveView process. In your case this message would be pushing message via websocket.

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -66,6 +67,37 @@ defmodule FormDemoLocal do
     end
   end
 
+  def handle_event(
```
kraleppa, line 70: Maybe you could create a new callback for server messages? Like `handle_server`?

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -66,6 +67,37 @@ defmodule FormDemoLocal do
...
+        "llv_server_message",
+        %{"type" => "synchronize", "users" => server_users},
+        socket
+      ) do
+    filtered_users =
+      Enum.filter(socket.assigns.users, fn user ->
+        case validate_already_existing(user, server_users) do
```
kraleppa, line 77: IMO validation if user already exists should happen on the server. If it was a real-life application you would probably store users in the database, and for me each validation that requires a database access should be handled on the server side. On submit you should send a message to the server, validate if user is unique, and return to LocalLiveView potential error message.

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -66,6 +67,37 @@ defmodule FormDemoLocal do
...
+          [] ->
+            send_to_phoenix(%{"type" => "new_user", "user" => user})
```
kraleppa, line 79: Similarly to sending a message from server to the browser - imo this should also be wrapped in some kind of function similar for LiveView users (like `send_update/3`). But this is a matter of crearing a proper API for LocalLiveViews

examples/form_demo/lib/form_demo_web/live/form_demo_live.ex
```diff
@@ -0,0 +1,61 @@
...
+  def handle_event(
+        "llv_local_message",
+        %{"payload" => %{"type" => "new_user", "user" => user}},
+        socket
+      ) do
+    new_users = socket.assigns.users ++ [user]
+    Application.put_env(FormDemo, :users, new_users)
+    socket = push_event(socket, "llv_rerender", %{"view" => "FormDemoLocal"})
+    {:noreply, assign(socket, users: new_users)}
+  end
```
kraleppa, line 35: Maybe a new callback for local_live_view messages? `handle_local`? So you could have:

```
def handle_local("new_user", %{"user" => user} = _params, socket) do
```

examples/form_demo/local/lib/form_demo_local.ex
```diff
@@ -30,14 +35,9 @@ defmodule FormDemoLocal do
 
   @impl true
   def mount(_params, _session, socket) do
-    users = [
-      %{"email" => "user1@example.com", "username" => "user1"},
-      %{"email" => "user2@example.com", "username" => "user2"},
-      %{"email" => "user3@example.com", "username" => "user3"}
-    ]
-
+    send(self(), :sync)
```
kraleppa, line 38: Not sure if I understand, but you are sending users from Server LiveView to Local LiveView, via 
```
 push_event(socket, "llv_server_message", %{"view" => "FormDemoLocal", "payload" => payload})
```

which is triggered just after `mount` of server LiveView. So why do you synchronize also on init of LocalLiveVIew?

kraleppa (commented):
Awesome stuff! I didn't really expect that we could exchange messages between Server LiveView and Local LiveView so fast! 

Imo the whole idea of sending events via websocket and then mapping them to popcorn events is correct. I'd say that is pretty elegant solution and doesn't break any LiveView conventions.

I made some API suggestions to make Local LiveViews a bit easier to develop and understand, but not sure if you want to introduce them now, since I guess you're in the PoC phase.

I think direct synchronization of states between Server LiveView and Local LiveView is not necessary, at least in this example. Keeping state in two places and synchronizing them is hard, also in real-life application you'd have to query the whole users table from the db, and send it to the browser. IMO the validation of user uniqueness should be handled on the purely on server side, and validate basic stuff like regex, and others in the browser

But I'd really like to see an example when you actually pass assigns via attributes to Local LiveView (I added a comment with small snippet visualizing how it could work), displaying this state, and maybe changing via form - this should probably trigger a message to the server, update the state of assigns on the server and trigger rerender of Local LiveView

examples/form_demo/lib/form_demo_web/live/form_demo_live.ex
FranekKubis: Thank you for suggestions - when it comes to API-like stuff - all your comments on the matter are valid, so I will prepare an issue to address them.

examples/form_demo/local/lib/form_demo_local.ex
FranekKubis: I agree the checks should be on the server. The behaviour was a leftover from the time when the FormDemo was a single LocalLiveView.
It was left as is for now because the demo should illustrate state-handling on the local side (so you can shut down the server - lost the user list - then start the serve and users will be synchronized from local to server).
I now its not the most similar to live behaviour but the point was to show the state sync not only server->local but also local->server.

examples/form_demo/local/lib/form_demo_local.ex
FranekKubis: I believe the part with trigerring sync from the server LiveView is a leftover - it did not work as expected because when the server LiveView mount was ready the local LiveView could not process the message yet.

examples/form_demo/lib/form_demo_web/live/form_demo_live.ex
FranekKubis: The passing assigns as attributes part seems understandable and even fairly simple to implement, but the whole synchronization mechanic between server and local LiveView assigns is a vast topic so this time also I would leave it as an issue for the future.

mat-hek (approved):

PR #445, opened 2025-12-23
Langtour restructure content

mmietla:
Restructure content of:

- Keyword lists and map
- Pattern matching
- Constructs
- Anonymous functions
- processes
- congratulations


misc/language-tour-guide/src/content/3-keyword-lists-and-maps/keyword_lists.livemd
```diff
@@ -0,0 +1,108 @@
...
+IO.inspect("hello", [label: "Print hello"])
+```
+
+You can pass multiple options to functions:
+
+```elixir
+#💡 Try removing the label, try changing order of keys
+#💡 Try using 'printable_limit' key to limit printed characters
+IO.inspect [{1, 2, 3}, "hello world"],
+  label: "Print a list", pretty: true, width: 15
```
mat-hek: add braces

mat-hek (approved):

PR #446, opened 2026-01-07
Fix module redefinition

mat-hek:
Our `code:is_loaded` patch calls `code:which`, which calls `code:is_loaded`, which leads to infinite recursion. This makes module redefinition hang (in many cases at least).

### DIFF
--- a/test/popcorn/eval_test.exs
+++ b/test/popcorn/eval_test.exs
@@ -199,7 +199,6 @@ defmodule Popcorn.EvalTest do
     |> AtomVM.assert_is_module()
   end
 
-  @tag :skip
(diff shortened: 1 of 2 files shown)


test/popcorn/eval_test.exs
```diff
@@ -199,7 +199,7 @@ defmodule Popcorn.EvalTest do
     |> AtomVM.assert_is_module()
   end
 
-  @tag :skip
+  @tag :module_redefinition
```
jgonet: Why we need the tag for that?

mat-hek: It's more convenient to run it if there's a tag, and I needed to run it a couple of times. Removed now

PR #449, opened 2026-01-13
Langtour: Capture failed editor code on atomvm crash

mmietla:


misc/language-tour-guide/src/utils/sentry/index.ts
```diff
@@ -40,14 +40,18 @@ export function captureReloadIframe(
...
+  const failedEditor = [...editors.values()].find(
+    (editor) => editor.executionState === "running"
+  );
+
+  const code = failedEditor
+    ? failedEditor.code
+    : editorOrder
+        .map((id) => {
+          const editor = editors.get(id);
+          return editor ? `# Editor: ${id}\n${editor.code}` : null;
```
jgonet: Maybe add executionState to id as well? Would be easier to find crashing one at glance.

PR #452, opened 2026-01-15
[Lang tour] Modules chapter

### LINKED ISSUE #457: [Lang tour] Modules chapter
- [x] Modules and functions
- [x] Import, alias, require
- [x] Module attributes, docs and typespecs
- [x] Behaviours and use

mat-hek:
closes #457

### DIFF
--- a/misc/language-tour-guide/src/content/7-modules/alias-require-import.livemd
+++ b/misc/language-tour-guide/src/content/7-modules/alias-require-import.livemd
@@ -2,7 +2,7 @@
 
 ## alias
 
-`alias` allows you to set up aliases for any given module name. It's useful when dealing with namespaced module names, like `MyProject.Utils`. The alias directive allows referring to it just as `Utils` within the current scope:
+`alias` allows you to refer to a module under a different name. It's useful when dealing with namespaced module names, like `MyProject.Utils`. The alias directive allows referring to it just as `Utils` within the current scope:
 
 ```elixir
 defmodule MyProject.Utils do
@@ -41,6 +41,12 @@ else
 end
 ```
 
+Finally, all module names and aliases are converted to atoms at the compile time:
+
+```elixir
+is_atom(Foo.Bar) and Foo.Bar == :"Elixir.Foo.Bar"
+```
+
 ## require
 
 Elixir provides macros as a mechanism for meta-programming (writing code that generates code). Macros are expanded at compile time.
@@ -63,6 +69,8 @@ require Logger, as: L
 
 L.info("hello")
 ```
+
+Note that `require` generates a _compile-time dependency_ on the required module. It means that the required module must be compiled before the one calling require. Also, if the required module changes, all modules which require it must be checked and may have to be recompiled. Therefore, using `require` too often may significantly increase compilation times.
 
 ## import
 
@@ -97,3 +105,36 @@ In the example above, the imported `List.duplicate/2` is only visible within tha
 While imports can be useful for frameworks and libraries to build abstractions, developers should generally prefer `alias` to `import` on their own codebases, as aliases make the origin of the function being invoked clearer.
 
 Also, note that `import` also `require`s given module automatically.
+
+
+## Multi alias, require and import
+
+Let's say we have multiple modules in the same namespace:
+
+```elixir
+defmodule My.Project do
+  defmodule Foo do
+    def hello do
+      "hello"
+    end
+  end
+
+  defmodule Bar do
+    def hi do
+      "hi"
+    end
+  end
+end
+```
+
+To alias them, we can use the syntax with `{}` brackets:
+
+```elixir
+alias My.Project.{Foo, Bar}
+
+Foo.hello() <> " " <> Bar.hi()
+```
+
+The same applies to `require` and `import`.
+
+💡 Try changing `alias` to `import` above. Adjust the calls accordingly.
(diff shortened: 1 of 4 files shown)


misc/language-tour-guide/src/content/7-modules/alias-require-import.livemd
```diff
@@ -0,0 +1,99 @@
+# Alias, require and import
+
+## alias
+
+`alias` allows you to set up aliases for any given module name. It's useful when dealing with namespaced module names, like `MyProject.Utils`. The alias directive allows referring to it just as `Utils` within the current scope:
```
jgonet: Maybe:

```suggestion
`alias` allows you to shorten or rename names of used modules. It's frequently used when dealing with namespaced module names, like `MyProject.Utils`. The `alias` directive allows referring to it just as `Utils` within the current scope:
```

misc/language-tour-guide/src/content/7-modules/alias-require-import.livemd
```diff
@@ -0,0 +1,99 @@
...
+end
+
+Foo.bar()
+```
+
+In the example above, the imported `List.duplicate/2` is only visible within that specific function. `duplicate/2` won't be available in any other function in that module (or any other module for that matter).
+
+While imports can be useful for frameworks and libraries to build abstractions, developers should generally prefer `alias` to `import` on their own codebases, as aliases make the origin of the function being invoked clearer.
+
+Also, note that `import` also `require`s given module automatically.
```
jgonet, line 107: I wonder if it's helpful to describe compile/runtime dependencies between modules somewhere here.

misc/language-tour-guide/src/content/7-modules/behaviours.livemd
```diff
@@ -0,0 +1,205 @@
...
+```elixir
+defmodule ProductPrinter2 do
+  # The product type for convenience
+  @type product :: %{name: String.t(), price: number()}
+
+  # @callback has the same syntax as @spec - there's just no implementation
+  @callback print(product) :: String.t()
+end
+
+defmodule PrettyProductPrinter2 do
```
jgonet: Maybe we should use some tags instead of numbers? `PrettyProductPrinterBehaviour`, `ProductPrinterWithDefault`, etc.

misc/language-tour-guide/src/content/7-modules/behaviours.livemd
```diff
@@ -0,0 +1,205 @@
...
+A behaviour is a module with _callback definitions_. Then, we can create modules that _implement_ the behaviour. Each of them should provide _callback implementations_, which are just functions. In our case, the behaviour needs one callback: `print/1`.
+
+```elixir
+defmodule ProductPrinter2 do
+  # The product type for convenience
+  @type product :: %{name: String.t(), price: number()}
+
+  # @callback has the same syntax as @spec - there's just no implementation
+  @callback print(product) :: String.t()
+end
```
jgonet: I'm wondering if we could make learning easier if we first presented the solution with two modules having the same API, without behaviours. Then, introduce library use-case (reading docs instead of having contract enforced by a compiler) and add the behavior module.

misc/language-tour-guide/src/content/7-modules/module-attributes.livemd
```diff
@@ -0,0 +1,166 @@
...
+  def circle_area_in_alternative_world(r) do
+    @pi * r ** 2
+  end
+end 
+
+# 💡 Try calling `Math2.circle_area(3)`. What's the result?
+Math2.circle_area_in_alternative_world(3)
+```
+
+However, note that the attributes are calculated during the compile-time: once the module is compiled, they can't change.
```
jgonet: IIRC they're also "pasted" – you can't get them after compilation and inferior to constant-like functions. Maybe worth noting.

misc/language-tour-guide/src/content/7-modules/module-attributes.livemd
```diff
@@ -0,0 +1,166 @@
+# Module attributes
+
+Module attributes in Elixir are constants computed at the compile-time that can be used within the given module:
```
jgonet, line 3: Maybe worth mentioning somwhere that some are for users and some are reserved and have meaning to the compiler.

misc/language-tour-guide/src/content/7-modules/modules-and-functions.livemd
```diff
@@ -0,0 +1,160 @@
...
+defmodule MyProject.Utils.Concat do
+  def join(a, b, sep \\ " ") do
+    a <> sep <> b
+  end
+end
+
+MyProject.Utils.Concat.join("hello", "world")
+```
+
+Actually, all modules are defined under `Elixir` namespace, but for convenience you can omit `Elixir.`
```
jgonet, line 122: Higher: "The module name should start with an uppercase letter."

Maybe worth adding some tidbit that you can create module with erlang-style name, nobody's stopping you.

I actually don't know if auto-prefix with `Elixir.` has some special meaning in compiler, splitting erlang and elixir modules for warnings for example.

misc/language-tour-guide/src/content/7-modules/alias-require-import.livemd
mat-hek: Hmm, that would mean you change the module name, while alias doesn't do that

jgonet: "You can use `alias` to refer to any module by different name." other idea.

My main issue is that having `alias` and "set up aliases" feels like this sentence doesn't provide any information to me.

PR #456, opened 2026-01-19
[Lang tour] Rephrase and extend the 'basic types' chapter, add follow-up changes elsewhere

mat-hek:


misc/language-tour-guide/src/content/1-basic-types-and-operations/atoms.livemd
```diff
@@ -31,7 +29,7 @@ is_atom(true)
 true == :true
 ```
 
-`IO.puts` returns `:ok`:
+Other commonly used atoms are `:ok` and `:error`. They usually denote a success or a failure of some operation. For example, `IO.puts/1` always returns `:ok`:
```
jgonet, line 32: I feel like explaining that `:ok`, `:error` can be used for status and then showing a function that hardcode success status is suboptimal :P

misc/language-tour-guide/src/content/1-basic-types-and-operations/basics.livemd
```diff
@@ -32,21 +32,27 @@ And a string:
...
-Similarly to many other languages, you may define a new variable by assigning a value to an identifier using `=` sign.
+We'll learn more about these types in a moment. Now, let's see how we can assign them to variables.
 
-In Elixir, all data is immutable. However, you may re-assign the variable name to a different value, even of a different type.
-
-Variables must start with a lowercase letter or an underscore. By convention, the variable names are written in `snake_case`. Leading underscore has a special meaning, but we'll get back to that later.
+## Variables and Immutability
 
-Assign a value to a variable:
+Similarly to many other languages, you can assign a value to a variable using `=` sign. Variables must start with a lowercase letter or an underscore. By convention, the variable names are written in `snake_case`. Leading underscore has a special meaning, but we'll get back to that later.
```
jgonet: Maybe worth linking a guide under "later". I hate when guides tell me "well, we'll get back to that" but don't tell me when or if it's advanced knowledge or just breaks the current flow explaining basics.

misc/language-tour-guide/src/content/1-basic-types-and-operations/basics.livemd
```diff
@@ -60,6 +66,37 @@ Re-assigning a variable doesn't affect other variables that were assigned its pr
...
-x = :ok
-{x, y}
+x = 1
+IO.inspect(x)
+y
+```
+
+## Modules and functions
+
+In Elixir, functions are grouped in modules. To call a function, we use `Module.function(comma_separated_arguments)`. We already called `inspect` function from the `IO` module:
```
jgonet: Idea: `Module.function(comma, separated, arguments)`

misc/language-tour-guide/src/content/1-basic-types-and-operations/booleans.livemd
```diff
@@ -20,13 +16,19 @@ Are they equal?
...
 ```elixir
 true and true
 ```
 
-Short-circuit `and`.
+```elixir
+true or false
+```
+
+Note that `and` and `or` are short-circuit: if the left-hand side determines the result, the right-hand side won't even be evaluated. It's especially important when doing side effects.
```
jgonet: Maybe: "It's especially important when used expressions have side-effects"

misc/language-tour-guide/src/content/1-basic-types-and-operations/booleans.livemd
```diff
@@ -90,3 +89,5 @@ The `!` operator:
 # 💡 Try `!1` and `!nil`.
 !true
 ```
+
+As a rule of thumb, use `and`, `or` and `not` when you are expecting booleans. If any of the arguments are non-boolean, use `&&`, `||` and `!`.
```
jgonet: Oxford comma, please.


I feel like this could be worded better, maybe along those lines:
> `&&`, `||`, and `!` should be used sparingly as they provide less information about expressions we're dealing with. Always use `and`, `or`, and `not` for boolean expressions.

misc/language-tour-guide/src/content/1-basic-types-and-operations/numbers.livemd
```diff
@@ -56,11 +54,13 @@ Exponential notation:
 1.0e-10
 ```
 
+Normally, integers can be arbitrarily big and they never overflow (or underflow). However, the AtomVM runtime which we're using in the browser only supports integers up to 64 bits and overflow raises an error. Floats are always stored on 64 bits and overflow raises an error.
```
jgonet, line 57: I feel like we should use admonitions like below for that type of info.

Image

misc/language-tour-guide/src/content/7-modules/module-attributes.livemd
```diff
@@ -40,7 +40,7 @@ defmodule Math2 do
   end
 end 
 
-# 💡 Try calling `Math2.circle_area(3)`. What's the result?
+# 💡 Try calling `Math2.circle_area/1`. What's the result?
```
jgonet, line 43: This changes slightly effort required for user.

mat-hek: That's the intent :D

misc/language-tour-guide/src/content/1-basic-types-and-operations/atoms.livemd
mat-hek: Agreed, but I can't find a function in stdlib that returns `:ok | :error`. Many return `:ok | {:error, reason}`, but we haven't introduced tuples yet. If you have a better idea, I'm all ears ;) BTW, we describe `{:ok, success} | {:error, failure}` case in the chapter about pattern matching.

misc/language-tour-guide/src/content/1-basic-types-and-operations/booleans.livemd
mat-hek: Hmm, I think the original version is clearer. It's copy-pasted from the Elixir guide, so I'm going to leave it. But I'll add a comma :D

misc/language-tour-guide/src/content/1-basic-types-and-operations/numbers.livemd
mat-hek: Agreed, but they currently don't work in the lang tour. Also, out of the scope of this PR

misc/language-tour-guide/src/content/1-basic-types-and-operations/atoms.livemd
jgonet: Then I'd link it here. I can't find any example as well. We could mention here that `:ok` is commonly used when function is doing only side-effects and we aren't interested in return value.

In any case, you can resolve without any change or do any change you think is helpful :P

Issue #444, opened 2025-12-19
[Lang tour] Scroll position jumps when running cells with pending outputs


mmietla:
When executing a cell further down the page without running previous cells first, the scroll position is not preserved. This occurs because the outputs from earlier cells are displayed simultaneously, causing the page layout to shift.

## Current fix
A partial fix has been implemented using `window.scrollBy()` to restore the scroll position based on editor top position. However, this approach still exhibits a visible flash where:
- The scroll position briefly jumps/shifts as outputs render

## Expected Behavior
When running any cell, the user's scroll position should remain stable without any visible jumping or flashing, even when outputs from previous cells are being rendered.

PR #462, opened 2026-01-22
Langtour: Prevent scroll position jumping when running editors

mmietla:
Implements: #444 

Added a toast notification when the code is completed and the editor is out of the user’s viewport

image

image

### DIFF
--- a/misc/language-tour-guide/src/components/Toast.tsx
+++ b/misc/language-tour-guide/src/components/Toast.tsx
@@ -1,26 +1,23 @@
 import { createPortal } from "react-dom";
-import type { ReactNode } from "react";
+import type { PropsWithChildren } from "react";
 import CloseIcon from "../assets/close.svg?react";
 
 type ToastProps = {
   visible: boolean;
   onClose?: () => void;
-  icon?: ReactNode;
   bgColor?: string;
   borderColor?: string;
   textColor?: string;
-  children: ReactNode;
 };
 
 export function Toast({
   visible,
   onClose,
-  icon,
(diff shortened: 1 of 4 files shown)


misc/language-tour-guide/src/components/Toast.tsx
```diff
@@ -0,0 +1,51 @@
...
+import CloseIcon from "../assets/close.svg?react";
+
+type ToastProps = {
+  visible: boolean;
+  onClose?: () => void;
+  icon?: ReactNode;
+  bgColor?: string;
+  borderColor?: string;
+  textColor?: string;
+  children: ReactNode;
```
jgonet: `PropsWithChildren`

misc/language-tour-guide/src/components/code-cell/EditorToast.tsx
```diff
@@ -0,0 +1,31 @@
...
+      bgColor="bg-light-30"
+      borderColor="border-grey-20"
+      textColor="text-brown-70"
+    >
+      <span className="font-inter text-sm font-medium">Code Completed</span>
+      <button
+        className={`text-brown-60 cursor-pointer text-xs`}
+        onClick={onClick}
+      >
+        Click to scroll
```
jgonet: Maybe `arrow-uturn-up` instead of `check-circle`, make both icon and "Code Completed" clickable and remove the text? I feel like the toast is kinda big and clickable region is near the close button region.

We could alternatively stack them vertically instead of horizontally.

Issue #373, opened 2025-10-27
[Lang tour] Implement E2E Tests


mmietla:
Implement end-to-end (E2E) tests to ensure that each topic within the lang tour works without errors.

PR #468, opened 2026-01-27
Langtour: E2E tests

mmietla:
Implements: #373 
Playwright tests run on all language tour pages, replacing code examples with correct versions from `.livemd` files. They execute all editors by clicking "Run Code" and verify execution state badges. The tests also check if there is no js error and console errors.

Changes needed to enable these tests:

- ### Navigation Structure Extraction

  - Extracted navigation config to separate module
  - Decoupled navigation data from Vite-specific module loading for testability
  - Added `getRoutes()`utility to generate all tour routes programmatically

- ### Correct Code Block Support
  - Enhanced .livemd parser to handle` langtour:{"test_replace_code":"code"}`  metadata comments
  - Added `data-correct-code` prop to CodeCell components for test automation

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -231,10 +231,10 @@ importers:
         version: 10.40.0(react@19.2.3)
       '@swmansion/popcorn':
         specifier: ^0.2.0
-        version: 0.2.0(esbuild@0.27.3)(rollup@4.55.2)(vite@7.3.1(@types/node@25.0.9)(jiti@2.6.1)(lightningcss@1.31.1))
+        version: 0.2.0(esbuild@0.27.3)(rollup@4.55.2)(vite@7.3.1(@types/node@24.12.0)(jiti@2.6.1)(lightningcss@1.31.1))
       '@tailwindcss/vite':
         specifier: ^4.1.11
-        version: 4.2.1(vite@7.3.1(@types/node@25.0.9)(jiti@2.6.1)(lightningcss@1.31.1))
+        version: 4.2.1(vite@7.3.1(@types/node@24.12.0)(jiti@2.6.1)(lightningcss@1.31.1))
       '@types/xxhashjs':
         specifier: ^0.2.4
         version: 0.2.4
@@ -293,6 +293,12 @@ importers:
       '@eslint/js':
         specifier: ^9.30.1
         version: 9.39.2
+      '@playwright/test':
+        specifier: ^1.56.1
+        version: 1.58.2
+      '@types/node':
+        specifier: ^24.9.1
+        version: 24.12.0
       '@types/react':
         specifier: ^19.1.9
         version: 19.2.9
@@ -301,7 +307,7 @@ importers:
         version: 19.2.3(@types/react@19.2.9)
       '@vitejs/plugin-react':
         specifier: ^4.6.0
-        version: 4.7.0(vite@7.3.1(@types/node@25.0.9)(jiti@2.6.1)(lightningcss@1.31.1))
+        version: 4.7.0(vite@7.3.1(@types/node@24.12.0)(jiti@2.6.1)(lightningcss@1.31.1))
       eslint:
         specifier: ^9.30.1
         version: 9.39.2(jiti@2.6.1)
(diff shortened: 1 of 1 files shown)


misc/language-tour-guide/src/content/4-pattern-matching/pattern_matching.livemd
```diff
@@ -56,6 +56,16 @@ IO.inspect(x, label: "x")
 IO.inspect(y, label: "y")
 ```
 
+<!-- langtour:{"correct_code":true} -->
```
mat-hek: Why is it only added in a few places? Can we assume the code is correct unless there's a `{"correct_code":false}` annotation?

mmietla: We can simply assume that code is correct if there is not a `{"correct_code":true}` annotation.
I added this annotation to allow adding correct_code from `.livemd` files, but the name may be misleading.

misc/language-tour-guide/src/content/8-processes/message_passing.livemd
```diff
@@ -100,3 +100,24 @@ end)
...
 Process.sleep(100)
 ```
+
+<!-- langtour:{"correct_code":true} -->
+
+```elixir
+p1 = spawn(fn ->
+  receive do
+    {:ping, pid} ->
+      IO.puts("recevied :ping, sending :pong")
```
jgonet: Typo.

misc/language-tour-guide/src/content/4-pattern-matching/pattern_matching.livemd
mat-hek: Ok, get it now! The name could be better, like (`testing-replace-code` ?) I also think the 'correct' version of the code should be placed within the comment, so it doesn't show up when you open it in Livebook. It could be moved before the code it replaces, too.

misc/language-tour-guide/src/content/5-constructs/if_and_cond.livemd
```diff
@@ -44,6 +44,17 @@ end
...
+<!-- langtour:{"correct_code":true} -->
+
+```elixir
+x = 1
+if true do
+  x = x + 1
+  x
+end
+x
+```
```
mat-hek: Why is this needed? Everything should be fine with the previous snippet 🤔

misc/language-tour-guide/src/content/8-processes/message_passing.livemd
```diff
@@ -100,3 +100,24 @@ end)
...
+
+p2 = spawn(fn ->
+  send(p1, {:ping, self()})
+  receive do
+    :pong -> IO.puts("received :pong")
+  end
+end)
+
+Process.sleep(100)
+```
```
mat-hek: Unless we want to use it to show the correct code to user somehow, I think this is not needed, as the original code doesn't fail

misc/language-tour-guide/src/content/5-constructs/if_and_cond.livemd
mmietla: It works fine, but it returns a warning on stderr, and the test failed because of that warning

misc/language-tour-guide/src/content/8-processes/message_passing.livemd
mmietla: Same problem as with the `if` example above

misc/language-tour-guide/src/content/5-constructs/if_and_cond.livemd
mat-hek: Hmm, I think warnings shouldn't make tests fail then, since it requires us to replace more code, and the replaced code isn't tested. But if it's problematic, it can stay as is.

misc/language-tour-guide/src/content/8-processes/message_passing.livemd
mat-hek: Ok, actually it's fine to have the full solution here anyway, to test it, so my original comment doesn't make sense

.github/workflows/langtour_e2e_test.yml
```diff
@@ -0,0 +1,69 @@
+name: Language Tour E2E tests
```
mat-hek: To be consistent with other workflow's names

```suggestion
name: Language Tour Build & test
```

PR #494, opened 2026-02-12
LocalLiveView: Integrate event handling with Phoenix

FranekKubis:
Changed custom JS handling to overwriting functions in View class objects.
It allows to use both localLiveViews and phoenixLiveViews at the same page.


examples/local-lv-compare/local/config/config.exs
```diff
@@ -3,6 +3,3 @@ import Config
 config :popcorn,
   out_dir: "../priv/static/local_live_view/wasm",
   add_tracing: false
```
mat-hek: this is the default

PR #561, opened 2026-03-25
js: replace bundlePath with bundlePaths array (support multi-bundle)

mmietla:
Add support for loading multiple .avm bundles in the Popcorn class. The `bundlePath` option is replaced by `bundlePaths: string[]`. Each bundle is served via a numbered meta tag (bundle-path-0, bundle-path-1, ...)

### DIFF
--- a/popcorn/elixir/test/fixtures/wasm/assets/index.js
+++ b/popcorn/elixir/test/fixtures/wasm/assets/index.js
@@ -7,7 +7,7 @@ window.createPopcornInstance = async function (bundlePath) {
   const logs = [];
 
   const popcorn = await Popcorn.init({
-    bundlePath,
+    bundlePaths: [bundlePath],
     onStdout: (output) => logs.push(`[popcorn stdout] ${output}\n`),
   });
 
--- a/popcorn/js/src/iframe.ts
+++ b/popcorn/js/src/iframe.ts
@@ -52,23 +52,17 @@ declare const globalThis: { TrackedValue: typeof TrackedValue };
 globalThis.TrackedValue = TrackedValue;
 
 export async function initVm(): Promise<void> {
-  const bundlePaths: string[] = [];
-  for (let i = 0; ; i++) {
-    const metaElement = document.querySelector(
-      `meta[name="bundle-path-${i}"]`,
-    ) as HTMLMetaElement | null;
-    if (!metaElement) break;
-    bundlePaths.push(metaElement.content);
-  }
-  if (bundlePaths.length === 0) {
-    throw new Error("Missing meta[name='bundle-path-0'] element");
-  }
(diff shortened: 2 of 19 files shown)


popcorn/elixir/test/fixtures/wasm/assets/index.js
```diff
@@ -1,9 +1,9 @@
 import { Popcorn } from "@swmansion/popcorn";
 
-const bundlePath =
-  new URLSearchParams(window.location.search).get("bundlePath") ?? undefined;
+const bundlePathParam =
+  new URLSearchParams(window.location.search).get("bundlePaths") ?? undefined;
```
jgonet: Maybe we can just use empty list for that?

popcorn/js/src/iframe.ts
```diff
@@ -51,19 +51,27 @@ class TrackedValue {
...
-    throw new Error("Missing meta[name='bundle-path'] element");
+export async function runIFrame(): Promise<void> {
+  const bundlePaths: string[] = [];
+  for (let i = 0; ; i++) {
+    const metaElement = document.querySelector(
+      `meta[name="bundle-path-${i}"]`,
+    ) as HTMLMetaElement | null;
+    if (!metaElement) break;
+    bundlePaths.push(metaElement.content);
   }
```
jgonet: this could be simpler:
```ts
const bundlePaths = Array.from(
  document.querySelectorAll<HTMLMetaElement>('meta[name^="bundle-path-"]'),
  el => el.content
);
```

popcorn/js/src/iframe.ts
```diff
@@ -51,19 +51,27 @@ class TrackedValue {
...
-  const bundlePath = metaElement.content;
-  const bundleBuffer = await fetch(bundlePath).then((resp) =>
-    resp.arrayBuffer(),
+  if (bundlePaths.length === 0) {
+    throw new Error("Missing meta[name='bundle-path-0'] element");
```
jgonet: I think we can just assert here, we should validate it's non-empty earlier.

popcorn/js/src/iframe.ts
```diff
@@ -51,19 +51,27 @@ class TrackedValue {
...
+  }
+  const bundles = await Promise.all(
+    bundlePaths.map((p) =>
+      fetch(p)
+        .then((resp) => resp.arrayBuffer())
+        .then((buf) => new Int8Array(buf)),
+    ),
```
jgonet: Maybe:
```js
  bundlePaths.map(async (p) => {
    const resp = await fetch(p);
    const buf = await resp.arrayBuffer();
    return new Int8Array(buf);
  })
```

original was to make it one-line-ish.

jgonet (approved):

PR #573, opened 2026-04-10
[LocalLiveView] LLV Generator

FranekKubis:
Tested on scenario:
1. `mix phx.new --no-ecto`
2. adding local_live_view to deps
3. `mix llv.install`
4. add local_live_view TestAppLocal template to home.html.heex
5. `mix setup`
6. `mix phx.server`


FranekKubis:
Changes to examples are simply just rewriting them using new mix llv.install command.
The implementation remains the same.

local-live-view/lib/mix/tasks/llv.build.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Mix.Tasks.Llv.Build do
```
mat-hek, line 1: Let's change to uppercase for all tasks
```suggestion
defmodule Mix.Tasks.LLV.Build do
```

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,471 @@
...
+
+      _ ->
+        content
+    end
+  end
+
+  defp inject_before_module_end(content, injection) do
+    [body | _] = Regex.split(~r/\nend\s*\z/, content)
+    body <> "\n\n" <> injection <> "end\n"
+  end
```
mat-hek: It looks like a heuristic that will sometimes work and sometimes not. If you want a generator that adds LLV to an existing project, it's better to use [ignitier](https://hexdocs.pm/igniter/readme.html)

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,471 @@
...
+        Mix.shell().info("* skipping #{path} esbuild (already configured)")
+      else
+        {ecto_in_setup, ecto_aliases} =
+          if String.contains?(content, "ecto.setup") do
+            {~s|"ecto.setup", |,
+             ~s|      "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],\n| <>
+               ~s|      "ecto.reset": ["ecto.drop", "ecto.setup"],\n|}
+          else
+            {"", ""}
+          end
```
mat-hek: How is Ecto related?

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,471 @@
...
+          content
+          |> String.replace(~r/  defp aliases do\n.*?\n  end\n/s, aliases_block)
+          |> inject_before_module_end("""
+            defp pnpm_install(_) do
+              {_, 0} =
+                System.cmd("pnpm", ["install"],
+                  cd: Path.join(File.cwd!(), "assets"),
+                  into: IO.stream(:stdio, :line),
+                  stderr_to_stdout: true
+                )
```
mat-hek: Is pnpm necessary to use popcorn/llv?

mat-hek (changes requested):
Generally, the generator looks very duct-taped, we can merge it so as to make it before the conference, but I think it should eventually be rewritten. Maybe we should try to limit the amount of generated code to make things simpler and more reliable.

Apart from that, the readme doesn't say how to use the task. I did `mix phx.new`, `llv.install`, `npm install`, `llv.build` and `mix phx.server` and got this error:

```
[notice] Application local_live_view exited: exited in: LocalLiveView.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (UndefinedFunctionError) function LocalLiveView.Application.start/2 is undefined (module LocalLiveView.Application is not available)
            LocalLiveView.Application.start(:normal, [])
            (kernel 9.0.2) application_master.erl:293: :application_master.start_it_old/4
[notice] Application telemetry exited: :stopped
[notice] Application popcorn exited: :stopped
[notice] Application jason exited: :stopped
[notice] Application decimal exited: :stopped
[notice] Application runtime_tools exited: :stopped
** (Mix) Could not start application local_live_view: exited in: LocalLiveView.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (UndefinedFunctionError) function LocalLiveView.Application.start/2 is undefined (module LocalLiveView.Application is not available)
            LocalLiveView.Application.start(:normal, [])
            (kernel 9.0.2) application_master.erl:293: :application_master.start_it_old/4
➜  my_llv_example git:(main)
```

EDIT: I missed the PR description. Why is `--dry-run` passed to `llv.install`? Anyway, following these steps gives me the following error (running `mix.setup`):
```
WARN  Moving esbuild that was installed by a different package manager to "node_modules/.ignored"
local_live_view is linked to /Users/matheksm/tmp/my_llv_example/assets/node_modules from /Users/matheksm/tmp/my_llv_example/deps/local_live_view/assets
 ERR_PNPM_LINKED_PKG_DIR_NOT_FOUND  Could not install from "/Users/matheksm/tmp/my_llv_example/deps/local_live_view/assets" as it does not exist.

This error happened while installing a direct dependency of /Users/matheksm/tmp/my_llv_example/assets
** (MatchError) no match of right hand side value: {%IO.Stream{device: :standard_io, raw: false, line_or_bytes: :line}, 1}
    mix.exs:86: MyLlvExample.MixProject.pnpm_install/1
    (mix 1.17.3) lib/mix/task.ex:574: Mix.Task.run_alias/6
    (mix 1.17.3) lib/mix/cli.ex:96: Mix.CLI.run_task/2
    /Users/matheksm/.asdf/installs/elixir/1.17.3/bin/mix:2: (file)
```

local-live-view/lib/mix/tasks/llv.install.ex
FranekKubis: it shouldnt be - fixed

local-live-view/lib/mix/tasks/llv.install.ex
FranekKubis: yes, also every example is using it

local-live-view/lib/mix/tasks/llv.build.ex
FranekKubis: Wouldnt the task change its name to l_l_v.build then? Seems like its better to leave it as it is.

local-live-view/lib/mix/tasks/llv.install.ex
FranekKubis: I changed the regexes to igniter usage.

local-live-view/lib/mix/tasks/llv.build.ex
FranekKubis: nvm it is possible to be not convention-dependend here

FranekKubis:
Overall I agree that the generator needs some fixes - as far as I could I change the approach to use Igniter, but it still can be written more cleanly.

I fixed obvious mistakes leading to generation errors, so it should work fine now outside of monorepo.

local-live-view/lib/mix/tasks/llv.build.ex
FranekKubis: nvm it is not possible xD
```
franek@MacBook-Pro-Franciszek test_app % mix llv.install

21:28:45.033 [info] Bundling started
[info] Bundling finished (took 18.29ms)
==> local_live_view
Compiling 6 files (.ex)
Generated local_live_view app
==> test_app
Compiling 14 files (.ex)
Generated test_app app
** (Mix) The task "llv.install" could not be found because the module is named Mix.Tasks.LLV.Install instead of Mix.Tasks.Llv.Install as expected. Please rename it and try again
```

FranekKubis:
As I wrote earlier - the most important changes are the addition of template files and llv.install and llv.build tasks.
The changes to examples are just to make sure they are regenerated by the llv.install mix task.

local-live-view/lib/mix/tasks/llv.build.ex
```diff
@@ -0,0 +1,32 @@
...
+      defp aliases do
+        [
+          setup: ["deps.get", "llv.build", ...]
+        ]
+      end
+  """
+
+  @impl Mix.Task
+  def run(_args) do
+    unless File.exists?("local") do
```
mat-hek:
```suggestion
    unless File.dir?("local") do
```

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,520 @@
...
+
+  defp find_import_by_suffix(zipper, suffix) do
+    case Igniter.Code.Function.move_to_function_call(zipper, :import, [1, 2], fn call ->
+           Igniter.Code.Function.argument_matches_predicate?(call, 0, fn arg ->
+             arg
+             |> Sourceror.Zipper.node()
+             |> Macro.to_string()
+             |> String.ends_with?(suffix)
+           end)
+         end) do
```
mat-hek: move this out of `case`

mat-hek (approved):

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,520 @@
...
+  end
+
+  # --- Application ---
+
+  defp inject_application(igniter) do
+    Igniter.Project.Application.add_new_child(
+      igniter,
+      {Registry, keys: :unique, name: LocalLiveView.ChannelRegistry}
+    )
+  end
```
mat-hek: This should be started by the liveview app

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,520 @@
...
+      [path | _] ->
+        Igniter.update_file(igniter, path, fn source ->
+          content = Rewrite.Source.get(source, :content)
+
+          if String.contains?(content, "local_live_view") do
+            source
+          else
+            llv_js =
+              ~s[\nimport { setup } from "vendor/local_live_view";\n] <>
+                ~s[setup(liveSocket, { Socket, bundlePath: "/assets/js/wasm/bundle.avm" });\n]
```
mat-hek: Very weird syntax. Verify the LLM-generated stuff for such things

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,520 @@
...
+              )
+            )
+          end
+        end)
+
+      [] ->
+        Igniter.add_warning(igniter, """
+        Could not find assets/js/app.js. Add after liveSocket.connect():
+            import { setup } from "vendor/local_live_view";
+            setup(liveSocket, { Socket, bundlePath: "/assets/js/wasm/bundle.avm" });
```
mat-hek: let's reuse the injected code instead of repeating it

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,520 @@
...
+    igniter
+    |> inject_endpoint()
+    |> inject_application()
+    |> inject_web_module()
+    |> inject_root_layout()
+    |> inject_app_js()
+    |> inject_package_json()
+    |> inject_mix_esbuild()
+    |> inject_dev_watcher()
+    |> generate_build_mjs()
```
mat-hek: The amount of injected things here is enormous. I know that the JS ecosystem is a mess, but we should really limit this. Any ideas @jgonet?

local-live-view/lib/mix/tasks/llv.install.ex
mat-hek, line 1: It would be useful to have a mix llv.new as well, a counterpart of mix phx.new. It could do
- mix phx.new
- add LLV to deps
- llv.install
- replace the default homepage with an LLV example

Just an idea, doesn't need to be added in this PR

mat-hek (changes requested):
Since it's still not merged after the conference, I believe we can polish it a bit more :P Most importantly, we should reduce the amount of JS-related boilerplate being generated

mat-hek:
Also, we should generate tests, and add a test to LLV that generates a basic project and checks if it works

local-live-view/lib/mix/tasks/llv.install.ex
jgonet: To be honest, IMO the approach should be "copy the template, interpolate changed things", we really need to only inject some mix tasks to build the JS, I don't know why we have so much code for JS there (didn't read it, looks LLM-generated).

jgonet:
One one-off comment: this PR is really big, would be nice to split it.

local-live-view/lib/mix/tasks/llv.install.ex
mat-hek: I believe the injection approach comes from the need to integrate with Phoenix

FranekKubis: Those are generated in order to change esbuild project config.
Esbuild in Phoenix does not allow additional JS code (popcorns' JS in our case) by default so thats why generator changes those files.

jgonet: Can you please collect these Phoenix limits? We could hand them to Phoenix maintainers, I think some general things like that are likely to be fixed.

mat-hek: Even with these Phoenix's limitations, I believe we can do better. Can't we just create a single bundle during LLV compilation and include it in `app.js`? Adding one line to `app.js` would then be the only thing the generator would have to do.

FranekKubis: I'll see what I can do here.

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -0,0 +1,306 @@
...
+  defp inject_esbuild_format(igniter) do
+    igniter
+    |> inject_esbuild_format_in("config/config.exs")
+    |> inject_esbuild_format_in("config/dev.exs")
+    |> inject_esbuild_format_in("config/prod.exs")
+  end
+
+  defp inject_esbuild_format_in(igniter, path) do
+    if File.exists?(path) do
+      replace_in_file(igniter, path, "--format=esm", "--bundle", "--bundle --format=esm")
```
mat-hek, line 231: Why do we bundle as ESM? Any ideas how to avoid it? cc @jgonet

local-live-view/lib/mix/tasks/llv.build.ex
```diff
@@ -0,0 +1,69 @@
...
+      Mix.raise("local/ directory not found. Run `mix llv.install` first.")
+    end
+
+    llv_path = Mix.Project.deps_paths()[:local_live_view]
+    assets_path = Path.join(llv_path, "assets")
+
+    Mix.shell().info("[llv] Installing JS dependencies...")
+
+    {_, 0} =
+      System.cmd(
```
mat-hek: NIT: the convention is to use `Mix.shell().cmd()`

local-live-view/lib/mix/tasks/llv.install.ex
FranekKubis: The problem with iife format is that in the iife format the line:
`const IFRAME_URL = new URL("./iframe.mjs", import.meta.url).href;`
is problematic, because import.meta.url is undefined. Popcorn expect JS to be in module format (esm).

mat-hek (approved):

PR #580, opened 2026-04-17
[LocalLiveView] Fixed thermostat and compare config

FranekKubis:
Added :wasm target and local_live_view dep to thermostat and compare examples.


examples/local-lv-compare/mix.exs
```diff
@@ -95,13 +96,7 @@ defmodule CompareLiveViews.MixProject do
...
 
   defp build_local(_) do
-    Mix.shell().cmd(
-      """
-      mix deps.get
-      mix popcorn.cook
-      """,
-      cd: "local"
-    )
+    Mix.shell().cmd("MIX_TARGET=wasm mix build", cd: "local")
```
mat-hek: Since the default target is WASM:

```suggestion
    Mix.shell().cmd("mix build", cd: "local")
```

mat-hek (approved):

mat-hek:
Disclaimer: I didn't check if it actually works :P

mmietla:
> Disclaimer: I didn't check if it actually works :P

I checked it, and it works fine.

PR #597, opened 2026-05-08
[LocalLiveView] Session bugfix

FranekKubis:
Fixed a bug where all LLVs were syncing to the same mirror.

### DIFF
--- a/examples/local-lv-compare/lib/compare_live_views_web/components/core_components.ex
+++ b/examples/local-lv-compare/lib/compare_live_views_web/components/core_components.ex
@@ -479,13 +479,25 @@ defmodule CompareLiveViewsWeb.CoreComponents do
 
   """
   attr :view, :string, required: true
-  attr :id, :string, doc: "stable element id, defaults to view name"
+  attr :id, :string, doc: "stable element id, defaults to server-generated random id"
 
   def local_live_view(assigns) do
-    assigns = assign_new(assigns, :id, fn -> assigns.view end)
+    assigns = assign_new(assigns, :id, fn -> Phoenix.LiveView.Utils.random_id() end)
+    assigns = assign(assigns, :has_mirror, mirror_exists?(assigns.view))
 
     ~H"""
-    <div data-pop-view={@view} id={@id} phx-update="ignore"></div>
+    <div
+      data-pop-view={@view}
+      id={@id}
+      data-pop-mirror={@has_mirror || nil}
+      phx-update="ignore"
+    >
+    </div>
     """
+  end
+
+  defp mirror_exists?(view_name) do
+    mirror = Module.concat(Mirror, String.to_atom(view_name))
+    Code.ensure_loaded?(mirror) and function_exported?(mirror, :handle_sync, 2)
   end
 end
--- a/examples/local-lv-compare/lib/compare_live_views_web/components/core_components.ex
+++ b/examples/local-lv-compare/lib/compare_live_views_web/components/core_components.ex
@@ -498,6 +498,6 @@ defmodule CompareLiveViewsWeb.CoreComponents do
 
   defp mirror_exists?(view_name) do
     mirror = Module.concat(Mirror, String.to_atom(view_name))
-    Code.ensure_loaded?(mirror) and function_exported?(mirror, :handle_sync, 2)
+    Code.ensure_loaded?(mirror) and function_exported?(mirror, :handle_sync, 3)
   end
 end
--- a/examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
+++ b/examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
@@ -4,7 +4,7 @@ defmodule FormDemoWeb.FormDemoLive do
   def render(assigns) do
     ~H"""
     <div class="centered-div">
-      <.local_live_view view="FormDemoLocal" />
+      <.local_live_view id={"form-demo-local-#{@socket.id}"} view="FormDemoLocal" />
       <div class="bordered">
         <h1>[Server Runtime] User List:</h1>
         <ul>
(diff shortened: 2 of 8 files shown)


local-live-view/lib/server/channel.ex
```diff
@@ -18,9 +32,16 @@ defmodule LocalLiveView.Channel do
     {:reply, socket.assigns.mirror_assigns, socket}
   end
 
+  def handle_call({:set_mirror_assigns, assigns}, _from, socket) do
+    push(socket, "set_assigns", assigns)
+    {:reply, :ok, socket}
+  end
+
   def handle_in("sync", local_assigns, socket) do
+    enriched = Map.put(local_assigns, "_llv_id", socket.assigns.llv_id)
```
mat-hek: It doesn't feel right to put the id in the assigns, I'd store it separately and add another argument to the mirror function (context? session?)

examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
```diff
@@ -18,9 +18,14 @@ defmodule FormDemoWeb.FormDemoLive do
   end
 
   def mount(_params, _session, socket) do
-    Phoenix.PubSub.subscribe(FormDemo.PubSub, "llv_mirror:FormDemoLocal")
-    mirror_assigns = LocalLiveView.Channel.get_mirror_assigns("FormDemoLocal")
-    {:ok, assign(socket, users: Map.get(mirror_assigns, "users", []))}
+    if connected?(socket) do
+      llv_id = "form-demo-local-#{socket.id}"
```
mat-hek, line 22: Isn't `socket.id` always set to `phx-`? See https://github.com/software-mansion/popcorn/blob/88239ea4302ee867b3bb24fd86b259bf966ce4bd/local-live-view/lib/local_live_view/server.ex#L385

examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
```diff
@@ -18,9 +18,14 @@ defmodule FormDemoWeb.FormDemoLive do
   end
 
   def mount(_params, _session, socket) do
```
mat-hek, line 20: Session is not used here, but as I see its id is `nil`.  Shouldn't it be set to something meaningful?

examples/local-lv-compare/lib/compare_live_views_web/components/core_components.ex
```diff
@@ -479,13 +479,25 @@ defmodule CompareLiveViewsWeb.CoreComponents do
...
+      phx-update="ignore"
+    >
+    </div>
     """
   end
+
+  defp mirror_exists?(view_name) do
+    mirror = Module.concat(Mirror, String.to_atom(view_name))
+    Code.ensure_loaded?(mirror) and function_exported?(mirror, :handle_sync, 2)
+  end
```
mat-hek, line 502: Can we make local_live_view function a `defdelegate` and move the implementation to the library?

local-live-view/lib/server/channel.ex
```diff
@@ -8,8 +8,22 @@ defmodule LocalLiveView.Channel do
     end
   end
 
+  def get_all_mirror_assigns(llv_id) do
+    Registry.select(LocalLiveView.ChannelRegistry, [
+      {{:"$1", :"$2", :"$3"}, [{:==, :"$3", llv_id}], [:"$2"]}
+    ])
+    |> Enum.map(&GenServer.call(&1, :get_mirror_assigns))
+  end
```
mat-hek: Is it for debug purposes?

mat-hek (changes requested):

examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
FranekKubis: No, socket.id in form_demo_live is an identifier of phoenix LiveView socket.
In LocalLiveView (e.g. form_demo_local) socket.id is set to "phx-" as it is not used.

examples/local-lv-compare/lib/compare_live_views_web/components/core_components.ex
FranekKubis: It is done in llv-generator PR.

examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
FranekKubis: I think socket.id is good enough for us - it survives reconnects and that is the demo purpose, so session can stay not used.

local-live-view/lib/server/channel.ex
FranekKubis: It is a leftover, I will remove it.

examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
mat-hek: Okay, I keep mistaking the server and the local part... Anyway, is it possible to get the liveview id  /  socket id / session id in the LLV? Do you think it would be useful to have access to any of them?

local-live-view/lib/server/channel.ex
```diff
@@ -8,8 +8,15 @@ defmodule LocalLiveView.Channel do
...
   end
 
+  def set_mirror_assigns(llv_id, assigns) do
+    case Registry.lookup(LocalLiveView.ChannelRegistry, llv_id) do
+      [{pid, _}] -> GenServer.call(pid, {:set_mirror_assigns, assigns})
+      [] -> {:error, :not_found}
+    end
+  end
+
   def join("llv:" <> llv_id, %{"view" => view_string}, socket) do
```
mat-hek, line 18: Since we receive llv_id from the client, isn't it possible to spoof it and change the state of someone else's liveview?

FranekKubis: Yes, it is possible - if someone does Man-in-the-Middle attack i guess or have someone elses browser access (but in that case nothing will help us xD).
I could add a signed token to the mirror sync mechanic. It will at least prevent brute-forcing and ID guessing.

examples/local-lv-forms/lib/form_demo_web/live/form_demo_live.ex
FranekKubis: Right now the access to "parent" Phoenix LiveView is not given out-of-the-box.
I think user should be responsible for sending it to LLV (eg rendering llv with "parent" socket.id).

local-live-view/lib/server/channel.ex
mat-hek: > It will at least prevent brute-forcing and ID guessing.

That's what I meant - it doesn't require MITM nor browser access.

> I could add a signed token to the mirror sync mechanic

Let's do that in a separate PR

mat-hek (approved):

PR #599, opened 2026-05-13
landing: add cookie consent proxy

p-malecki:
* Added a local Cookie Script proxy serving /cookie-script.js from the same origi
* Injected the locally served Cookie Script into both the language tour and landing page


jgonet:
Done in #605.

PR #600, opened 2026-05-14
Add experimental tree-shaking

mat-hek:
Tree-shaking reduces the generated bundle (.avm) size by removing unused functions. For the `local thermostat` example, it reduces the bundle size from ~4,6MB (compressed) to ~1MB (compressed). Unfortunately, tree-shaking for Elixir can't be 100% accurate, and it can remove functions that may be used. Manual adjustment is possible, but it's not a public API (yet). Details are described in the `Treeshake` module.

As a by-product, I changed non-patched modules so that they're not passed through our recompilation pipeline, reducing the compilation speed. For some reason, the kernel module stopped working because of that, so I added an empty patch for it.

I also refactored the `Popcorn` module, simplifying it a bit:
- It no longer creates separate `.avm`s for each stdlib app
- The generated boot module is smaller - limited to what needs to be dynamically generated
- `Popcorn.cook`s can now be run in parallel

Vibe-coded modules are marked in the comment at the top of the module.

Todo:
- [x] Test treeshake in Popcorn and HelloPopcorn


.github/workflows/popcorn_treeshake_test.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "24"
+          cache: "pnpm"
+          cache-dependency-path: pnpm-lock.yaml
```
jgonet, line 53: Why can't we use the composite action instead of copy & pasting the steps?

examples/local-lv-thermostat/local/config/config.exs
```diff
@@ -2,3 +2,5 @@ import Config
 
 config :popcorn,
   out_dir: "../priv/static/assets/js/wasm"
+
+config :popcorn, treeshake: true
```
jgonet: Other check the env var, is it correct to hardcode here?

popcorn/elixir/lib/popcorn/build.ex
```diff
@@ -97,77 +98,50 @@ defmodule Popcorn.Build do
 
     elixir_apps = Path.expand("..", Application.app_dir(:elixir)) |> File.ls!()
 
-    otp_apps ++ elixir_apps
```
jgonet, line 100: If we don't use `String.to_existing_atom`, please add a comment about that so we know it wasn't accidental.

popcorn/elixir/lib/popcorn/build.ex
```diff
@@ -97,77 +98,50 @@ defmodule Popcorn.Build do
...
 
-  defp patchable_app_beams(:popcorn_lib), do: []
+  defp patchable_app_ebins(:popcorn_lib), do: []
 
-  defp patchable_app_beams(app) do
+  defp patchable_app_ebins(app) do
     app
     |> Application.app_dir()
-    |> Path.join("ebin/**/*.beam")
+    |> Path.join("ebin/*")
```
jgonet: I assume the intent is to capture .app files, shouldn't we do that explicitly instead of blanket wildcard?

popcorn/elixir/lib/popcorn/build.ex
```diff
@@ -250,37 +225,42 @@ defmodule Popcorn.Build do
   # but we execute it always for consistency
   # To be replaced with File.cp when we improve tracing in AtomVM
   defp do_patch(app_name, name, nil, patch, out_dir) do
-    Logger.info("Patching #{name}", app_name: app_name)
+    Logger.info("Transferring #{name}", app_name: app_name)
```
jgonet, line 230: I'm not sure if this is accurate if we conditionally patch it here.

popcorn/elixir/lib/treeshake/utils/beam_analyzer.ex
```diff
@@ -0,0 +1,357 @@
+defmodule Treeshake.Utils.BeamAnalyzer do
+  @moduledoc false
+  # Extracts information about functions, behaviours and protocols from core erlang
+  # Mostly vibe-coded
```
jgonet, line 4: Could use some cleanup but looks reasonable for first glance (isn't that case for all LLM output, though?).

popcorn/elixir/test/fixtures/treeshake/ebin/Elixir.Task.beam
jgonet, line 1: Can we use some lighter beam?

popcorn/elixir/test/fixtures/treeshake/call_graph.bin
jgonet, line 1: How do we use it?

popcorn/elixir/test/popcorn/compilation_test.exs
jgonet, line 40: I think I saw that one failing at some point. Maybe we can rework it so it still checks stacktraces?

popcorn/elixir/test/treeshake/utils/beam_rewriter_test.exs
```diff
@@ -0,0 +1,157 @@
...
+      ]
+
+      {:ok, :NoDebugInfoMod, binary, _} =
+        :compile.forms(forms, [:return_errors, :return_warnings])
+
+      tmp =
+        Path.join(System.tmp_dir!(), "no_debug_info_#{:erlang.unique_integer([:positive])}.beam")
+
+      File.write!(tmp, binary)
+      on_exit(fn -> File.rm(tmp) end)
```
jgonet: We can use tmp dirs from exunit I think.

.github/workflows/popcorn_treeshake_test.yml
mat-hek: What's a composite action? Treeshake tests are separated because they take a while, and there's no point in running them unless treeshake logic has changed

popcorn/elixir/lib/popcorn/build.ex
mat-hek: I'd say it's a matter of taste, I can make it explicit if you prefer

popcorn/elixir/lib/popcorn/build.ex
mat-hek: By patching we usually mean applying a patch from popcorn patches, here we at most inject the tracing stuff

popcorn/elixir/lib/treeshake/utils/beam_analyzer.ex
mat-hek: Yes, I added these comments to mark the code I only briefly reviewed, and so should you :P Which is fair IMO, considering treeshake's experimental state :P

popcorn/elixir/test/fixtures/treeshake/call_graph.bin
mat-hek: There's a snapshot test that fails when it changes

popcorn/elixir/test/popcorn/compilation_test.exs
mat-hek: It's a leftower, will unskip

popcorn/elixir/test/fixtures/treeshake/ebin/Elixir.Task.beam
mat-hek: It's 38K, is it that much?

.github/workflows/popcorn_treeshake_test.yml
jgonet: Ah, nevermind. I thought we have one action like `Setup toolchain` for all pnpm and other needed things but I see some are unmigrated. Let's keep it this way for now, until I get back (again) for meddling with CI.

popcorn/elixir/lib/popcorn/build.ex
jgonet: Whatever works, I have preference for explicitness since I don't know what gets put into ebin on top of my head (and explicit extensions suggest that nicely).

popcorn/elixir/test/fixtures/treeshake/ebin/Elixir.Task.beam
jgonet: It isn't very bad, but I try really hard to avoid putting binary files in the repo. Can stay this way though.

popcorn/elixir/test/fixtures/treeshake/call_graph.bin
mat-hek: I compressed it so it's smaller now

popcorn/elixir/test/fixtures/treeshake/ebin/Elixir.Task.beam
mat-hek: It can actually be taken from the system - removed

jgonet (approved):

FranekKubis (approved):

PR #603, opened 2026-05-14
[LocalLiveView] add pushEvent API for calling from external JS

mmietla:


FranekKubis:
Please check if it still works after I merged: https://github.com/software-mansion/popcorn/pull/597.

mmietla:
> Please check if it still works after I merged: #597.

It works well

local-live-view/assets/local_live_view.js
```diff
@@ -189,6 +194,25 @@ export async function setup(liveSocket, opts = {}) {
   for (const detail of bufferedServerMessages) {
     await sendServerMessage(popcorn, detail);
   }
+
```
FranekKubis, line 222: I dont really like the idea of returning an object containing our API, by the setup function.
I was thinking about it and Iin my opinion if we are adding the pushEvent functionality we should change the export API a bit:
instead of exporting setup function we should export a class (named e.g., LLVEngine or something like this)
This new class should be doing everything setup function does in the contructor and provide pushEvent as a method.
something like this:

```
export class LocalLiveView {
  constructor(liveSocket, config) {
    this.liveSocket = liveSocket;
    this.config = config;
    this.popcorn = null;
    this.isReady = false;

    this.init();
  }

  async init() {
    // setup functionality
  }

  async pushEvent(viewId, event, payload = {}) {
    if (!this.isReady) {
       console.warn("LLV not ready yet");
       return;
    }
    return await this.popcorn.call(...);
  }
}
```

FranekKubis (changes requested):

examples/local-lv-compare/assets/js/app.js
```diff
@@ -39,14 +39,10 @@ topbar.config({ barColors: { 0: "#29d" }, shadowColor: "rgba(0, 0, 0, .3)" });
 window.addEventListener("phx:page-loading-start", (_info) => topbar.show(300));
 window.addEventListener("phx:page-loading-stop", (_info) => topbar.hide());
 
-// connect if there are any LiveViews on the page
-liveSocket.connect();
```
FranekKubis, line 43: Why are we removing liveSocket.connect? As far as I remember we need it to use phoenix JSes correctly.
Did you test the solution?

mmietla: `liveSocket.connect` was moved to `LLVEngine.create`

mmietla: I tested it on the landing page and standalone

FranekKubis: Another problem is that when you create a new phoenix project it already does liveSocket.connect() in app.js (so another .connect would lead to unspecified consequences).
The llv.install on top of it should only modify the liveSocket interfaces - shouldnt really connect the socket itself.

I would revert it.

FranekKubis (changes requested):

FranekKubis (approved):

PR #608, opened 2026-05-21
[LLV] Pong demo

mat-hek:
The first commit is a result of `mix phx.new` and `mix llv.install`, the second contains the actual changes, the third contains small improvements to LLV. The game logic is vibe-coded.

Edit: Also added timer support to Popcorn via `timer_manager` and changed pong to use it (Claude originally used timer, originally I had to change it as it wasn't supported).

Todo:
- [x] display current FPS
- [x] make target FPS configurable


local-live-view/lib/local_live_view/local_live_view.ex
```diff
@@ -50,6 +50,13 @@ defmodule LocalLiveView do
     socket
   end
 
+  @doc """
+  Mimics `Phoenix.LiveView.connected?/1`. Always returns `true`.
+
+  Helps code reusability between server and local LiveViews.
+  """
+  def connected?(_socket), do: true
```
mat-hek, line 58: Not sure if it makes sense, but now in many cases one can do

```elixir
defmodule ModuleLive do
  if Mix.target() == :wasm do
    use LocalLiveView
  else
    use Phoenix.LiveView
  end
  ...
```

and have a reusable liveview

FranekKubis: I dont think "helping code reusability" is the only thing this place does so I would leave it as it is now.

examples/local-lv-pong/README.md
```diff
@@ -0,0 +1,18 @@
+# LocalLvPong
```
FranekKubis: Maybe a quick description here.

popcorn/elixir/patches/popcorn_lib/eavmlib/timer_manager.erl
```diff
@@ -175,6 +180,24 @@ run_timer(MgrPid, Time, TimerRef, Dest, Msg) ->
         Dest ! {timeout, TimerRef, Msg}
     end.
 
+%% @private
+run_interval(MgrPid, Time, _TimerRef, Dest, Msg) ->
+    Start = erlang:system_time(millisecond),
+    do_run_interval(MgrPid, Start, Time, 0, Dest, Msg).
+
+%% @private
+do_run_interval(MgrPid, Start, Time, Iteration, Dest, Msg) ->
```
FranekKubis, line 190: If Dest dies, the timer stays and this function will loop infinitely, at scale leading to memory leak (timer stays uncancelled and alive), am I right?

mat-hek: Good catch, should be fixed

FranekKubis (approved):

PR #617, opened 2026-05-25
popdoc: initial version

jgonet:
Popdoc is an ExDoc plugin which adds Popcorn integration in order to make code-blocks interactive. Thanks for mmietla for initial version that solidified this approach.

Features:
- line-by-line execution – last result shown by default, section that expands with all results. Only visible if we have multiple expressions.
- stacktraces – trimmed to user code, hidden by default (only error with message is shown). Only visible if there's an error.
- easy setup – one hook for ExDoc, all artifacts precompiled. We create bundle of user code when generating docs to allow using modules from library that generated docs. To make codeblock interactive, add `<!-- popcorn:eval -->` comment.
- mobile-friendly.

To test locally (may need some mix deps.get in popdoc dir):
in `<repo>/popdoc/e2e/fixture`
```bash
mix popdoc.build
mix deps.get && mix docs && mix popdoc.server
```

Follow-ups:
- `iex>` handling
- publish the package
- lazy loading of .avm and VM

https://github.com/user-attachments/assets/fc8dc291-93b3-42d8-8acd-9b88f0623706


popdoc/README.md
mmietla, line 1: Missing docs on how to make a code block executable, add a `<!-- popcorn:eval -->` example

mmietla (approved):

PR #618, opened 2026-05-25
popcorn: add abstraction over cleanup functions

### LINKED ISSUE #188: Design API for cleanup functions
Some refs need a function to specify how they should be cleaned up. For example, when creating an interval, we should remove it when it's no longer needed (process needing interval died, ref linked with interval was dropped, we want to remove it manually).

## Current solution
We generate key and use low-level map on Wasm module to store those functions. Excerpt from code:
```js
// `node` is DOM node, `event_name` is name (e.g. "click"), `fn` is event callback, `wasm` is emscripten's wasm object
node.addEventListener(event_name, fn);

const key = wasm.nextTrackedObjectKey();
const cleanupFn = () => {
    node.removeEventListener(event_name, fn);
    wasm.cleanupFunctions.delete(key);
};

wasm.cleanupFunctions.set(key, cleanupFn);
return [new TrackedValue({key: key, value: cleanupFn})];
```

It uses `wasm.nextTrackedObjectKey()`, `wasm.cleanupFunctions` key -> function/0 map, and `TrackedValue` to prevent another key from being generated.

## Solution

We should allow refs to have associated cleanup functions without having to resort to creating key manually and using key->destructor mapping directly.

> [!NOTE]
> TODO: propose some solutions

jgonet:
Some approach to #188.
Closes #188.

### DIFF
--- a/popcorn/elixir/lib/popcorn/api/wasm.ex
+++ b/popcorn/elixir/lib/popcorn/api/wasm.ex
@@ -329,14 +329,7 @@ defmodule Popcorn.Wasm do
       };
 
       node.addEventListener(event_name, fn);
-      const key = wasm.nextTrackedObjectKey();
-      const cleanupFn = () => {
-        node.removeEventListener(event_name, fn);
-        wasm.cleanupFunctions.delete(key);
-      };
-      wasm.cleanupFunctions.set(key, cleanupFn);
-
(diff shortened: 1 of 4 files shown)


mmietla (approved):

jgonet:
Closing since we won't need that in the future.

PR #634, opened 2026-06-08
[LLV] Kanban demo

mat-hek:
Demo showcasing simple forms, modals, and drag&drop with LLV. Server sync will come in another PR.

Other changes:
- Support for mouse and drag JS events in LLV - I'm not sure if it's the best way
- Support for `phx-target` for live components - again, not sure if this is how it should be done
- Clone objects received from Popcorn - fixes a weird bug when a diff wouldn't render
- Fix `maps:take` return value - found accidentally, didn't break anything so far AFAIK


local-live-view/assets/local_live_view.js
```diff
@@ -117,7 +124,15 @@ export class LLVEngine {
         );
         return;
       }
-      view.update(diff, []);
+      // The diff is delivered from the Popcorn WASM iframe via run_js, so its
```
FranekKubis, line 127: Isn't the 'realm problem' a problem for every time when we want to deliver an array over a popcorns run_js?
Shouldn't the solution be handled on the "realm-border" as a part of popcorn functionality?

local-live-view/lib/local_live_view/server.ex
```diff
@@ -118,6 +132,59 @@ defmodule LocalLiveView.Server do
...
+  end
+
+  defp component_handle_event_call(component_socket, _component, "lv:clear-flash", val) do
+    case val do
+      %{"key" => key} -> Utils.clear_flash(component_socket, key)
+      _ -> Utils.clear_flash(component_socket)
+    end
+  end
+
+  defp component_handle_event_call(_component_socket, _component, "lv:" <> _ = bad_event, _val) do
```
FranekKubis: This seems similar to view_handle_event function - maybe we can refactor this to DRY up the code.

local-live-view/lib/local_live_view/server.ex
```diff
@@ -118,6 +132,59 @@ defmodule LocalLiveView.Server do
...
+    case result do
+      {diff, new_components, _extra} ->
+        {:noreply, push_diff(%{state | components: new_components}, diff, ref)}
+
+      :error ->
+        {:noreply, push_noop(state, ref)}
+    end
+  end
+
+  defp component_handle_event_call(component_socket, _component, "lv:clear-flash", val) do
```
FranekKubis: This seems similar to view_handle_event function - maybe we can refactor this to DRY up the code.

local-live-view/lib/server/watcher.ex
```diff
@@ -33,7 +33,7 @@ defmodule LocalLiveView.Watcher do
 
   @spec start_link([option]) :: GenServer.on_start()
   def start_link(opts \\ []) do
-    opts = Keyword.validate!(opts, dirs: [Path.absname("local")])
+    opts = Keyword.validate!(opts, dirs: [Path.absname("local/lib")])
```
FranekKubis: What is the purpose of this change?
Also we need to keep in mind this will influence other examples.

examples/local-lv-kanban/local/lib/local/column_component.ex
```diff
@@ -0,0 +1,77 @@
...
+  attr :drag_target, :map, default: nil
+
+  def column(assigns) do
+    assigns = assign(assigns, :tasks, OrderedMap.values(assigns.col.tasks))
+
+    ~H"""
+    <div
+      phx-dragover="drag_over_column"
+      phx-value-column_id={@col.id}
+      style={"flex:0 0 280px;background:#1f2937;border:1px solid #{if @dragging, do: "#4b6bb3", else: "#374151"};border-radius:8px;padding:0.75em;display:flex;flex-direction:column;gap:0.6em"}
```
FranekKubis, line 22: For me inline-style'ing is suboptimal when it comes to readability.

local-live-view/lib/server/watcher.ex
mat-hek: it was a leftover ;)

local-live-view/assets/local_live_view.js
mat-hek: I'll do that in another PR

examples/local-lv-kanban/local/lib/local/column_component.ex
mat-hek: I'll do it in the PR with backend integration, as it also changes the client part a bit

FranekKubis (approved):

PR #635, opened 2026-06-09
[LLV] Refactor JS distribution

FranekKubis:
## Summary

Refactor JS distribution: ship `local_live_view.js` pre-bundled inside the hex package instead of building it user-side via npm. End users now import via package name (`import "local_live_view"`) instead of relative vendor path, and `mix llv.build` drops the npm/node steps entirely.

## Changes

**Library side:**
- New `local-live-view/package.json` with `main: "priv/static/local_live_view.js"` for esbuild's NODE_PATH resolution (NODE_PATH points to "deps" directory).
- `priv/static/{local_live_view.js, iframe.mjs, AtomVM.mjs, AtomVM.wasm}` committed (pre-bundled at release time via `assets/bundle.mjs`).
- `mix llv.build` simplified: copies runtime files from `deps/local_live_view/priv/static/` to user's `priv/static/assets/js/`, then builds WASM. No npm, no node.
- `llv.install` template injects `import { LLVEngine } from "local_live_view"` (was `from "../vendor/local_live_view.js"`).

**Examples** (`local-lv-{compare,forms,thermostat,pong,burrito}`):
- Migrated to standard Phoenix esbuild pipeline (away from custom `build.mjs` + node watcher).
- (!) Each `assets/package.json` declares `"local_live_view": "workspace:*"` so pnpm symlinks LLV into `node_modules/` (path deps don't get `deps/<name>/` from Mix — workspace fills the gap in monorepo).
- `bundlePaths` fixed to match `out_dir` (`/assets/js/wasm/bundle.avm`).
- `pnpm-workspace.yaml` includes `local-live-view` and missing example entries.


local-live-view/priv/static/AtomVM.mjs
mat-hek, line 1: Shouldn't we gitignore everything in priv/static?

FranekKubis: As far as I know lots of libraries (phoenix included) is storing and maintaining this kindof regenerated files inside repository for the ease of using the whole library as a dependency.

mat-hek (approved):

PR #636, opened 2026-06-11
[LLV-burrito] Add info modal, llv loader and UX improvements 

mmietla:
image

image


examples/local-lv-burrito/assets/js/app.js
```diff
@@ -266,6 +266,19 @@ window.toggleElevator = function () {
...
+};
+
+window.closeInfoModal = function () {
+  document.getElementById("info-modal").classList.add("hidden");
+  localStorage.setItem("llv-burrito-seen", "1");
+};
+
+if (!localStorage.getItem("llv-burrito-seen")) {
+  document.getElementById("info-modal")?.classList.remove("hidden");
+}
```
mat-hek: This looks terribly hacky, I'd either
- wrap this in a LiveView's JS hook and explicitly use it in the modal component
- write everything in vanilla JS, including the modal itself
- add another LLV that manages the modal and accesses local storage via `Popcorn.Wasm.run_js`

Claude should be able to implement either solution rather autonomously

mat-hek (approved):

PR #637, opened 2026-06-11
landing: add topbar banner

p-malecki:
Adds a topbar banner to the landing page. Content is served server-side via external service, so banners can be swapped, rotated, or pulled without any repo change or redeploy. Banners are currently hidden.


jgonet:
Superseded by #750.

PR #641, opened 2026-06-19
[LLV] Assigns & server events

mat-hek:
In this PR:

- support for passing assigns to LLV
- support for pushing events to the server
- rewrite of kanban demo to use assigns
- persisting kanbans in a db, support for multi-user preview


local-live-view/lib/local_live_view/server.ex
```diff
@@ -169,6 +184,32 @@ defmodule LocalLiveView.Server do
...
+    end
+  end
+
+  # Top-level assign keys cross the JSON boundary as strings; convert them back to
+  # atoms so they read like Phoenix assigns (`@items`). Nested values are left as
+  # is — deeply atomizing arbitrary maps would be unsafe.
+  defp normalize_assigns(assigns) do
+    Map.new(assigns, fn
+      {key, value} when is_atom(key) -> {key, value}
+      {key, value} when is_binary(key) -> {String.to_atom(key), value}
```
FranekKubis, line 209: `String.to_existing_atom/1` ? + unknown key handling maybe

local-live-view/assets/local_live_view.js
```diff
@@ -368,6 +438,32 @@ export class LLVEngine {
...
+    // LLV's host LiveView fresh each call (a reconnect can swap the
+    // [data-phx-session] element) and dispatches the event to it over the
+    // websocket.
+    window.__llvPushServer = (llvId, event, payload) => {
+      const popEl = document.getElementById(llvId);
+      if (!popEl) {
+        console.error("LLV pushServer: LLV element not found", llvId);
+        return;
+      }
+      const hostEl = popEl.closest(PHX_VIEW_SELECTOR);
```
FranekKubis: Maybe we can combine host resolution and error handling here to some `withHost(el, (hostView, hostEl) => …)`. Then use it here and in line 59.

local-live-view/lib/local_live_view/server.ex
mat-hek: I actually forgot about this - I think this is generally suboptimal and we should use `term_to_binary` and `binary_to_term` instead of JSON for stuff that never touches JS. I'd do that in a separate PR.

For now, I wouldn't change it - since it's run on the client, atom exhaustion is not a problem and `to_existing_atom` would only make the error more cryptic - `ArgumentError` vs unknown assign passed to LLV. Unknown key handling - how would we know which key is 'known'?

FranekKubis (approved):

local-live-view/lib/local_live_view/server.ex
FranekKubis: I meant that 'known' key would be an existing atom.

PR #643, opened 2026-06-23
Deserialize run_js args using main window's realm

mat-hek:


jgonet (approved):
Can we get some tests for that?

popcorn/elixir/lib/popcorn/api/wasm.ex
```diff
@@ -368,14 +368,23 @@ defmodule Popcorn.Wasm do
...
+      # `args.key instanceof Array` also works inside `run_js`.
+      # This unfortunately makes literals not work: `[] instanceof Array`
+      # and `{} instanceof Object` inside `run_js` returns `false`.
       code = """
       (Module) => {
         const iframeWindow = globalThis.window;
         const window = globalThis.window.parent;
         const document = globalThis.window.parent.document;
+        const Array = window.Array;
+        const Object = window.Object;
```
jgonet, line 384: I'm afraid this is not enough in the general case, should work for the moment though.

mat-hek: JSON.parse only generates objects and arrays, so it's not possible to have other entities, if that's what you mean

PR #654, opened 2026-07-02
[LLV] Add token-based security to LocalLiveView mirrors

### LINKED ISSUE #647: [LLV] Verify IDs received from the client
Currently, in the LLV mirror we receive an LLV id from the client and don't verify it: https://github.com/software-mansion/popcorn/blob/main/local-live-view/lib/server/channel.ex#L18

We should probably sign off these IDs to prevent enumeration/spoofing and check make sure there are no other places like this.

mmietla:
closes: #647

Replace boolean mirror checks with signed tokens verified on channel join.
Add CSRF protection to WebSocket connection by passing session to connect_info.
Tokens are signed with the Phoenix endpoint, preventing unauthorized access
to mirror components.


local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -74,9 +74,16 @@ defmodule Mix.Tasks.Llv.Install do
              use_zipper,
              """
 
+             @session_options [
+               store: :cookie,
+               key: "_app_key",
+               signing_salt: "salt"
+             ]
```
mat-hek: This seems to be already generated by Phoenix

local-live-view/lib/server/channel.ex
```diff
@@ -15,10 +15,27 @@ defmodule LocalLiveView.Channel do
...
+      {:ok, %{id: ^llv_id, view: ^view_string}} ->
+        case Registry.register(LocalLiveView.ChannelRegistry, llv_id, view_string) do
+          {:ok, _} ->
+            mirror_module = find_mirror_module(view_string)
+
+            {:ok,
+             assign(socket, llv_id: llv_id, mirror_assigns: %{}, mirror_module: mirror_module)}
+
+          {:error, {:already_registered, _pid}} ->
+            {:error, %{reason: "already_joined"}}
```
mat-hek: In the way the mirrors work (which I just learned) it's expected that many LLVs can share the same mirror

local-live-view/lib/server/channel.ex
```diff
@@ -45,11 +62,13 @@ defmodule LocalLiveView.Channel do
...
 
   defp find_mirror_module(view_string) do
-    mirror = Module.concat(Mirror, view_string)
+    mirror = String.to_existing_atom("Elixir.Mirror." <> view_string)
 
     if Code.ensure_loaded?(mirror) and function_exported?(mirror, :handle_sync, 3),
       do: mirror,
       else: nil
+  rescue
+    ArgumentError -> nil
```
mat-hek:
```suggestion
    mirror =
      try do
        String.to_existing_atom("Elixir.Mirror." <> view_string)
      rescue
        ArgumentError -> nil
      end

    if mirror != nil and Code.ensure_loaded?(mirror) and function_exported?(mirror, :handle_sync, 3) do
      mirror
    else
      nil
    end
```

local-live-view/lib/server/component.ex
```diff
@@ -63,7 +67,9 @@ defmodule LocalLiveView.Component do
   defp default_id(name), do: "llv-" <> String.replace(name, ~r/[^A-Za-z0-9_-]/, "-")
 
   defp mirror_exists?(view_name) do
-    mirror = Module.concat(Mirror, String.to_atom(view_name))
+    mirror = String.to_existing_atom("Elixir.Mirror." <> view_name)
     Code.ensure_loaded?(mirror) and function_exported?(mirror, :handle_sync, 3)
+  rescue
+    ArgumentError -> false
   end
```
mat-hek: Looks like it's repeated in the channel, let's unify it

local-live-view/lib/server/mirror_token.ex
```diff
@@ -0,0 +1,29 @@
+defmodule LocalLiveView.MirrorToken do
+  @moduledoc """
```
mat-hek: This module should probably be private

mat-hek (changes requested):

local-live-view/lib/server/channel.ex
FranekKubis: While it is possible via overwriting the id of the LLV component, we can discuss whether it should be possible or not.
By default the LLV component is created with random generated id.

examples/local-lv-burrito/lib/burrito_web/live/order_live.ex
```diff
@@ -108,7 +108,7 @@ defmodule BurritoWeb.OrderLive do
               </div>
               <span class="text-xs text-pop-orange-dark/60 font-medium">Loading Local LiveView…</span>
             </div>
-            <.local_live_view id={@llv_id} view="BurritoLive" />
```
FranekKubis, line 111: I guess we would like the assigns.endpoint to always be socket.endpoint (at least by default). Am I right?

mmietla: yes, but the user should explicitly pass it because we can't access the `socket.endpoint` from the `local_live_view` component.

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -67,40 +67,49 @@ defmodule Mix.Tasks.Llv.Install do
...
+             ),
+           {:ok, session_opts_zipper} <- find_attr(zipper_with_plug, :session_options) do
+        {:ok,
+         Igniter.Code.Common.add_code(
+           session_opts_zipper,
+           """
+           socket "/llv_socket", LocalLiveView.Socket,
+             websocket: [connect_info: [session: @session_options]]
+           """,
+           placement: :after
```
mat-hek: These two operations (adding headers and passing session options) are independent, so when one fails the other can still succeed. Let's at least move them to separate private functions

mat-hek (approved):

FranekKubis (approved):

PR #656, opened 2026-07-07
[LLV] Restructure local-live-view package

mmietla:
Move package.json to root and consolidate build setup. Use popcorn esbuild plugin directly instead of manually copying runtime files. Update workspace references in mise.toml and pnpm lockfile accordingly.


local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -238,6 +239,45 @@ defmodule Mix.Tasks.Llv.Install do
     end
   end
 
+  # --- esbuild alias for path deps ---
+
+  # Phoenix's esbuild resolves the bare `import "local_live_view"` in app.js via
+  # NODE_PATH=deps, but Mix does not materialize path deps under deps/ — for a
+  # path dep, point esbuild directly at the dep's committed JS bundle instead.
+  # The alias resolves the dep path when the config is evaluated (the ~w sigil
+  # interpolates), so the generated config stays machine-independent.
```
jgonet: There's no `~w` sigil there, this is wrong and redundant info. Shouldn't this be just:
> For local paths, Mix creates a symlink which is not included in deps/. Alias imports to local_live_view with path to generated JS code.
> https://esbuild.github.io/api/#alias

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -238,6 +239,45 @@ defmodule Mix.Tasks.Llv.Install do
...
+      |> inject_esbuild_alias_in("config/dev.exs")
+      |> inject_esbuild_alias_in("config/prod.exs")
+    else
+      igniter
+    end
+  end
+
+  defp llv_path_dep? do
+    llv_path = Mix.Project.deps_paths()[:local_live_view]
+    llv_path != nil and Path.expand(llv_path) != Path.expand("deps/local_live_view")
```
jgonet, line 295: I don't know what this is exactly checking.

local-live-view/assets/bundle.mjs
```diff
@@ -29,18 +28,9 @@ const buildOpts = {
   bundle: true,
   format: "esm",
   outfile: resolve(out, "local_live_view.js"),
+  plugins: [popcorn({ bundlePaths: [] })],
```
jgonet, line 32: Maybe worth a comment why we don't pass bundle paths.

examples/local-lv-forms/config/config.exs
```diff
@@ -35,7 +35,7 @@ config :esbuild,
   version: "0.25.4",
   form_demo: [
     args:
-      ~w(js/app.js --bundle --format=esm --target=es2022 --outdir=../priv/static/assets/js --external:/fonts/* --external:/images/* --alias:@=.),
+      ~w(js/app.js --bundle --format=esm --alias:local_live_view=#{Mix.Project.deps_paths()[:local_live_view]}/priv/static/local_live_view.js --target=es2022 --outdir=../priv/static/assets/js --external:/fonts/* --external:/images/* --alias:@=.),
```
jgonet: Would be great to maybe create a helper to know why this alias is needed. Also, please group aliases together (there's one as last arg).

local-live-view/lib/mix/tasks/llv.install.ex
mat-hek: Mix doesn't create any symlinks for local paths

jgonet: Right, it copies? In any case, this comment should be explaining why we're do this in fewer and more accurate words.

mat-hek: No, it just reads from that path. I think the only redundant sentence is about the `~w` sigil, which is in another place.

mat-hek:
@mmietla please use popcorn v0.3.1 in LLV

jgonet (approved):

PR #667, opened 2026-07-13
Patch Base.encode64 and decode64 to rely on Erlang's base64 module

mat-hek:


jgonet (approved):
I'd add few tests with different encoded strings so we don't break edge cases.

PR #674, opened 2026-07-14
[LLV] Restructure local-live-view as modular TypeScript

mmietla:
Module Restructuring
- Split local_live_view.js into modules:
  - events.ts — Event handling and dispatching
  - helpers.ts — Utility functions
  - index.ts — Main entry point and view initialization
  - navigation.ts — Navigation and routing
  - types.ts — Shared type definitions
  - view_setup.ts — fake view setup

Build Tooling
- Added Rollup configuration with TypeScript support
- ESLint configuration with TypeScript rules
- Prettier formatting configuration
- Generated .d.ts type definitions for the bundle


local-live-view/assets/local_live_view/core/view_registry.ts
mat-hek, line 1: this looks like a wrapper over a map, do we really need it?

local-live-view/assets/local_live_view/core/popcorn_client.ts
mat-hek, line 1: why is this in a `core` subdir?

local-live-view/assets/local_live_view/events.ts
```diff
@@ -0,0 +1,127 @@
...
+      },
+    };
+  };
+
+  // Walk up from `target` to the nearest element that carries the `binding`
+  // attribute (e.g. "phx-dragover"), returning it or null if none is found.
+  // Mirrors how phx-click resolves a handler from the event's deepest target.
+  const closestWithBinding = (target: EventTarget | null, binding: string): Element | null => {
+    let el = target as Node | null;
+    while (el && el.nodeType === 1 && !(el as Element).getAttribute?.(binding)) {
```
jgonet: [Node.ELEMENT_NODE](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#node.element_node) instead of `1`. We can write narrowing function so we don't need `as Element` cast.

local-live-view/assets/local_live_view/index.ts
```diff
@@ -0,0 +1,288 @@
...
+      },
+      destroyed() {
+        unmountView(this.el);
+      },
+    } satisfies Hook;
+  }
+
+  // Pages with only LocalLiveViews (no server-side LiveView) connect in "dead"
+  // mode, which skips bindForms() — making phx-submit / phx-change no-ops on
+  // any LLV. Wire them up manually when no real LiveView is on the page.
```
jgonet, line 218: There are a lot of comments here, do we need all of them?

local-live-view/assets/local_live_view/core/popcorn_client.ts
jgonet: Could be inlined in the module that uses it tbh.

local-live-view/priv/static/AtomVM.mjs
jgonet, line 1: Why we're including generated artifacts in the repo?

local-live-view/priv/static/AtomVM.wasm
jgonet, line 1: Ditto, this is binary file.

local-live-view/assets/local_live_view/index.ts
mat-hek: Not sure if all, but many of them address LV quirks and are quite helpful. I'd say it's not in scope of this refactor anyway.

local-live-view/priv/static/AtomVM.mjs
mat-hek: IIRC it's because we deliver assets the same way as Phoenix does, cc @FranekKubis

FranekKubis: Exactly – we keep these files in `priv` to simplify the setup and avoid rebuilds during dependency compilation. Phoenix does it the same way.

local-live-view/assets/local_live_view/core/view_registry.ts
mmietla: I want to add an abstraction for managing the views. In the future, we might want to add methods for iterating through them, but i agree that it could be an overengineered appraoch

mat-hek (approved):

PR #678, opened 2026-07-17
[LLV] Share llv version and code between local and host part of the llv apps

### LINKED ISSUE #611: [LLV] Improve local project compilation strategy
Issues with the current approach
- [x] Popcorn and LLV are compiled twice.
- [x] LLV versions may diverge between local and server projects. Same for potential dependencies that would ship both local and server parts.
- [x] The server project doesn't have access to the local project, which could enable safe code reuse.

FranekKubis:
Closes: https://github.com/software-mansion/popcorn/issues/611

The host app now shares its `deps/` and `mix.lock` with the `local` (WASM) project via `deps_path`/`lockfile`, and `local/mix.exs` carries a `Mix.target()` gate (`elixirc_paths`/`application`) so it's a no-op on the host build while still compiling for wasm through `popcorn.cook`. `mix llv.install` was updated to generate exactly this setup — a gated `local/mix.exs` plus a `{:local, path: "local"}` dependency added to the host. 

**Note** that the `local_live_view` version is still declared independently in each `mix.exs`, so keeping the two in sync remains the user's responsibility — the generator only guarantees a matching version at install time.


FranekKubis:
Why not an umbrella

We deliberately keep the host (Phoenix/BEAM) app and the `local` (WASM) project as two plain Mix projects — host with `local/` as a subdirectory, wired via `{:local, path: "local"}` — rather than an umbrella. Reasons:

- **We don't want to force users into an umbrella.** Adopting LocalLiveView shouldn't dictate the top-level project layout. Most Phoenix apps are plain single projects; requiring an umbrella (or a migration to one) is a large, invasive prerequisite for what should be an additive feature.

- **Umbrella forces a shared config, but ours are disjoint.** An umbrella centralizes configuration and shares one `MIX_ENV`/`MIX_TARGET` across all apps. The host and the local view have genuinely disjoint config: the host carries the Phoenix/endpoint/esbuild/tailwind config (host target only), while `local` carries Popcorn config such as the WASM `out_dir` (wasm target only). Merging them into one shared config is a mismatch — e.g. the host build then sees `config :popcorn`, an app it doesn't depend on, and warns.

- **The forced shared config makes us fight our own target-based compilation.** `mix compile` / `mix phx.server` at an umbrella root compile *every* app under a single (host) target. But `local` is wasm-only — its views `use LocalLiveView`, whose main module only compiles for the wasm target (it references Popcorn, which is `targets: :wasm`). To survive the umbrella's host-wide compile we had to add contortions: target-gate `elixirc_paths`/`application` so `local` compiles to nothing on host, `reloadable_apps` so the code reloader skips it, and `config_target` guards around the Popcorn config. None of that is needed once `local` isn't an umbrella member.

- **On top of that, umbrella adds a per-app directory level**, so every phx.new tool path baked with `../deps` (esbuild `NODE_PATH`, heroicons vendor, etc.) has to be manually deepened. More tax for no benefit here.

The two things an umbrella would have bought us — a single shared version and code reuse between host and local — we get without it: the shared dep spec via one declaration point (`local/mix.exs`, pulled transitively into the host), and reuse via ordinary target-neutral modules. So the umbrella's only advantages don't require the umbrella.

mat-hek (approved):
Looks good, we should probably update all examples

mat-hek:
> Note that the local_live_view version is still declared independently in each mix.exs, so keeping the two in sync remains the user's responsibility — the generator only guarantees a matching version at install time.

@FranekKubis I think it's not possible for them to diverge, since the host project depends on the local one. In fact, we should be able to specify the LLV dependency in the local project only, as the host project gets it transitively.

FranekKubis:
> > Note that the local_live_view version is still declared independently in each mix.exs, so keeping the two in sync remains the user's responsibility — the generator only guarantees a matching version at install time.
> 
> @FranekKubis I think it's not possible for them to diverge, since the host project depends on the local one. In fact, we should be able to specify the LLV dependency in the local project only, as the host project gets it transitively.

We still need :local_live_view dep in the main project because we need `mix llv.install` task which then creates `local` directory during boilerplate generation.

mat-hek:
Ok, so in the future we can encourage having `{:local_live_view, ">= 0.0.0"}` in one project and the actual constraint in the other

PR #695, opened 2026-07-23
popdoc: split popdoc.js into src/ modules

mmietla:
> Part 1/5 of the popdoc IEx integration

Adds runnable IEx snippets to ExDoc pages. Each prompt in ` ```iex-popcorn ` fenced blocks becomes clickable and replays in a floating in-browser IEx terminal backed by Popcorn.

## How it works

- **Build-time parsing (`Popdoc.Markdown`)**: `iex-popcorn` fences are parsed during `mix docs` (same pipeline as `popcorn-eval`). Command list is embedded as JSON in a `data-popcorn-iex-commands` attribute.
- **Evaluation (wasm, `PopdocWasm.Eval`)**: stateless eval helpers shared by eval blocks and the terminal; the `PopdocWasm` GenServer holds the IEx session's binding + env.
- **Terminal UI**: floating xterm.js panel (Clear / Reset / Collapse / Close, theme synced with ExDoc's light/dark toggle) plus an `iex` launcher button. Clicking a snippet replays its commands sequentially with running / success / failure indicators; a failure stops the chain.
- **Lazy boot**: runtime boots on first interaction (snippet click or launcher), not on page load.

## Decisions

- **Prebuild parsing**: commands extracted at `mix docs` time and shipped in an HTML data attribute, mirroring the `popcorn-eval` approach.
- **xterm.js over ghostty-web**: ghostty had focus handling bugs; xterm is stable.
- **Simple eval over porting Popcorn's IEx example terminal**: in the example, the IEx `Shell` module writes output straight to the terminal via JS calls from Elixir, so the JS side never sees a structured result. There's no way to sync a command with its outcome, which makes waiting indicators and error handling per snippet hard to implement. Instead the terminal is a thin JS shell that sends each line to Elixir for plain evaluation and gets the result back as data.


popdoc/js/src/popdoc.js
```diff
@@ -0,0 +1,122 @@
...
+let popcornInitPromise = null;
+
+// Live instance shared with every module. Reset swaps it, so consumers read
+// it at call time instead of capturing it in closures at bind time.
+export function getPopcorn() {
+  return popcornInstance;
+}
+// Bumped per exdoc:loaded so auto block ids never collide across SPA
+// navigations — the VM (and its per-block sessions) outlives the page.
+let pageEpoch = 0;
```
jgonet: This wasn't in the original file?

jgonet (changes requested):

popdoc/js/src/popdoc.js
mmietla: It wasn't in the original file

jgonet: It's way easier to review if we split logic changes from code moves. Can we please split the code first and then in the next PR change the logic?

mmietla: yeap

jgonet (commented):
lgtm, would be a bit easier if this was additionally split into two commits (one moving files, non-compiling; second updating imports).

jgonet (approved):

PR #696, opened 2026-07-23
popdoc: add IEx session to wasm module (2/5)

mmietla:
> Part 2/5 of the popdoc IEx integration. Part of #695

### DIFF
--- a/popdoc/wasm/lib/wasm.ex
+++ b/popdoc/wasm/lib/wasm.ex
@@ -3,10 +3,12 @@ defmodule PopdocWasm do
 
   import Popcorn.Wasm, only: [is_wasm_message: 1]
   alias Popcorn.Wasm
+  alias PopdocWasm.Eval
 
   @process_name :main
   @ellipsis_sentinel :__popdoc_ellipsis__
   @snippet_max_len 40
+  @syntax_colors IO.ANSI.syntax_colors()
 
   def start_link(args) do
     GenServer.start_link(__MODULE__, args, name: @process_name)
@@ -14,10 +16,12 @@ defmodule PopdocWasm do
 
   @impl true
   def init(_args) do
+    # Evaluated user code runs in this process; a crashing linked process
+    # (spawn_link, Task.async) must deliver an exit message instead of
+    # killing the session with all its bindings.
+    Process.flag(:trap_exit, true)
     Wasm.ready(@process_name)
-    # TODO: needed?
-    :application.set_env(:elixir, :ansi_enabled, false)
-    {:ok, %{sessions: %{}}}
+    {:ok, %{sessions: %{}, iex: nil}}
   end
 
   @impl true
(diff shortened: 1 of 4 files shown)


popdoc/wasm/lib/iex_session.ex
```diff
@@ -0,0 +1,131 @@
+defmodule PopdocWasm.IexSession do
```
jgonet: I kinda don't see point of this module. What it actually does? I feel like it adds a struct we don't need (we already have one for binding and env from what I remember and could only be `PopcornWasm.Eval` with eval functions).

popdoc/wasm/lib/iex_session.ex
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  # Re-parse with string_to_quoted! for the exception's full message
+  # (SyntaxError/MismatchedDelimiterError render richer text than the raw
+  # error tuple).
+  defp parse_error(code) do
+    Code.string_to_quoted!(code)
+  rescue
+    err -> exception_to_error_map(err, "")
+  end
```
jgonet: What about:

```
defp parse_input(code) do
  {:ok, Code.string_to_quoted!(code)}
rescue
  _ in TokenMissingError -> :incomplete
  err -> {:error, exception_to_error_map(err, "")}
end
```
and simplify eval?

jgonet (changes requested):

popdoc/wasm/lib/iex_session.ex
mmietla: First I want to create an `IexSession` module to separate the session logic from the evaluation logic. Since both currently share some functionality, it would be cleaner to rename the existing module to PopcornEval and keep only the eval logic there imo.

popdoc/wasm/lib/wasm.ex
```diff
@@ -72,12 +76,45 @@ defmodule PopdocWasm do
     end
   end
 
+  # Sent on SPA navigation: eval-block sessions never outlive their page, and
+  # a run that spans the navigation gets a clean "no active session" reject.
```
jgonet, line 80: " and  # a run that spans the navigation gets a clean "no active session" reject."

What does that mean 😭

popdoc/wasm/lib/wasm.ex
```diff
@@ -3,21 +3,25 @@ defmodule PopdocWasm do
...
   def start_link(args) do
     GenServer.start_link(__MODULE__, args, name: @process_name)
   end
 
   @impl true
   def init(_args) do
+    # Evaluated user code runs in this process; a crashing linked process
+    # (spawn_link, Task.async) must deliver an exit message instead of
+    # killing the session with all its bindings.
+    Process.flag(:trap_exit, true)
```
jgonet, line 22: Maybe we should run code under [task supervisor](https://elixir.hexdocs.pm/Task.Supervisor.html#async_nolink/2?) instead? Docs [don't recommend](https://elixir.hexdocs.pm/Task.html#async/1-linking:~:text=In%20any%20case,caller%20process%20dies.) setting flags.

PR #697, opened 2026-07-23
popdoc: parse iex-popcorn blocks in markdown (3/5)

mmietla:
> Part 3/5 of the popdoc IEx integration. Part of #695


popdoc/lib/popdoc/markdown.ex
```diff
@@ -60,6 +88,54 @@ defmodule Popdoc.Markdown do
...
+    final = if current != nil, do: [current | commands], else: commands
+    Enum.reverse(final)
+  end
+
+  defp strip_prompt(line, prompt_re) do
+    case Regex.run(prompt_re, line) do
+      [prompt] -> String.replace_prefix(line, prompt, "")
+      nil -> nil
+    end
+  end
```
jgonet, line 127: This feels like it could be `Regex.replace` if we change logic a bit and remove nil (which could improve rest of the code).

popdoc/lib/popdoc/markdown.ex
```diff
@@ -60,6 +88,54 @@ defmodule Popdoc.Markdown do
...
+
+  # ExDoc.Utils.to_json escapes every control character — a raw byte < 0x20
+  # in a doc line would otherwise make JSON.parse throw and silently disable
+  # the block. ExDoc is guaranteed loaded here: to_ast/2 already requires its
+  # Earmark adapter.
+  defp commands_to_json(commands) do
+    Module.concat([ExDoc, Utils])
+    |> apply(:to_json, [commands])
```
jgonet, line 135: Don't we depend on ExDoc?

popdoc/lib/popdoc/markdown.ex
```diff
@@ -60,6 +88,54 @@ defmodule Popdoc.Markdown do
...
+          code = strip_prompt(trimmed, @iex_prompt_re) ->
+            acc = if current != nil, do: [current | commands], else: commands
+            {acc, code}
+
+          (cont = strip_prompt(trimmed, @cont_prompt_re)) && current != nil ->
+            {commands, current <> "\n" <> cont}
+
+          true ->
+            {commands, current}
+        end
```
jgonet, line 115: This needs to change, assignments in cond, we use `&&`, and a lot of nils everywhere. Most likely this can be rewritten by tagging strip_prompt with different atoms.

popdoc/lib/popdoc/markdown.ex
```diff
@@ -26,12 +32,27 @@ defmodule Popdoc.Markdown do
...
+        iex_code_block?(code_attrs) ->
+          commands_json =
+            code_children |> extract_text() |> parse_iex_commands() |> commands_to_json()
+
+          new_pre_attrs =
+            pre_attrs
+            |> add_class(@iex_class)
+            |> Keyword.put(:"data-popcorn-iex-commands", commands_json)
+
+          {new_pre_attrs, normalize_code_attrs(code_attrs)}
```
jgonet, line 49: This feels like it mixes a lot of abstraction. What about:
- `type(code_attrs) -> :iex | :code_block | :other`
- `add_data(attrs, key, value)`
- call `normalize_code_attrs` after the cond/case block (maybe with rename to `dedup()`)

PR #702, opened 2026-07-24
[LLV] Figure out how to assign IDs to LLVs

### LINKED ISSUE #679: [LLV] Figure out how to assign IDs to LLVs
Changing LLV's id to be deterministic fixed problems with the id changing between the two LV mounts, but caused conflicts in LLV mirrors.

FranekKubis:
Closes: https://github.com/software-mansion/popcorn/issues/679

Summary:

- LLV ID handling: Retained the default behavior. If a single instance of a LocalLiveView is present, `id` defaults to `"llv-LLVModuleName"`. If multiple instances of the same LocalLiveView exist, the user must explicitly provide a unique `id` for each instance.

- Mirror identification: The server uses `mirror_id` to uniquely identify mirrors connected to a specific LLV instance. The `mirror_id` is a combination of `socket.id` and the component's `id`.

### DIFF
--- a/examples/local-lv-burrito/lib/burrito_web/live/sync_table_live.ex
+++ b/examples/local-lv-burrito/lib/burrito_web/live/sync_table_live.ex
@@ -16,12 +16,20 @@ defmodule BurritoWeb.Live.SyncTableLive do
     "guacamole" => 2.50
   }
 
-  def mount(_params, %{"llv_id" => llv_id}, socket) do
-    if connected?(socket) do
-      Phoenix.PubSub.subscribe(Burrito.PubSub, "llv_mirror:BurritoLive:#{llv_id}")
-    end
+  def mount(_params, %{"llv_rendering_socket_id" => llv_rendering_socket_id}, socket) do
+    attrs =
+      case connected?(socket) do
+        true ->
+          mirror_id =
+            LocalLiveView.Component.mirror_id(llv_rendering_socket_id, "llv-BurritoLive")
 
-    attrs = if connected?(socket), do: LocalLiveView.Channel.get_mirror_assigns(llv_id), else: %{}
+          Phoenix.PubSub.subscribe(Burrito.PubSub, "llv_mirror:BurritoLive:#{mirror_id}")
+          LocalLiveView.Channel.get_mirror_assigns(mirror_id)
+
+        false ->
+          %{}
+      end
+
     {:ok, apply_attrs(socket, attrs)}
   end
 
--- a/examples/local-lv-burrito/lib/burrito_web/live/sync_table_live.ex
+++ b/examples/local-lv-burrito/lib/burrito_web/live/sync_table_live.ex
@@ -18,16 +18,14 @@ defmodule BurritoWeb.Live.SyncTableLive do
 
   def mount(_params, %{"llv_rendering_socket_id" => llv_rendering_socket_id}, socket) do
     attrs =
-      case connected?(socket) do
-        true ->
-          mirror_id =
-            LocalLiveView.Component.mirror_id(llv_rendering_socket_id, "llv-BurritoLive")
+      if connected?(socket) do
+        mirror_id =
+          LocalLiveView.Component.mirror_id(llv_rendering_socket_id, "llv-BurritoLive")
 
-          Phoenix.PubSub.subscribe(Burrito.PubSub, "llv_mirror:BurritoLive:#{mirror_id}")
-          LocalLiveView.Channel.get_mirror_assigns(mirror_id)
-
-        false ->
-          %{}
+        Phoenix.PubSub.subscribe(Burrito.PubSub, "llv_mirror:BurritoLive:#{mirror_id}")
+        LocalLiveView.Channel.get_mirror_assigns(mirror_id)
+      else
+        %{}
       end
 
     {:ok, apply_attrs(socket, attrs)}
--- a/local-live-view/assets/local_live_view/index.ts
+++ b/local-live-view/assets/local_live_view/index.ts
@@ -29,6 +29,7 @@ interface CreateArgs {
   url: string;
   urlParams: Record<string, string>;
   assigns: string;
+  mirrorId: string | undefined;
 }
 
 class PopcornClient {
@@ -58,7 +59,7 @@ class PopcornClient {
     );
   }
 
-  create({ id, view, url, urlParams, assigns }: CreateArgs): Promise<CallResult> {
+  create({ id, view, url, urlParams, assigns, mirrorId }: CreateArgs): Promise<CallResult> {
     return this.call({
       action: "create",
       id,
@@ -66,6 +67,7 @@ class PopcornClient {
       url,
       url_params: urlParams,
       assigns,
+      mirror_id: mirrorId,
     });
   }
 
@@ -75,14 +77,6 @@ class PopcornClient {
 
   reconnected(id: string): void {
     this.fire("reconnect sync", { action: "reconnected", id, payload: {} });
-  }
-
-  connectMirror(id: string, mirror_id: string): void {
-    this.fire("connect mirror", {
-      action: "connect_mirror",
-      id,
-      payload: { mirror_id: mirror_id },
-    });
   }
 
   updateAssigns(id: string, assigns: string): void {
@@ -190,6 +184,8 @@ export class LLVEngine {
   // Start a view and wire it up.
   private async mountView(pop_view_el: HTMLElement): Promise<void> {
     const llvId = pop_view_el.id;
+    const mirrorId = pop_view_el.dataset.popMirrorId;
+    this.maybeSetupMirrorChannel(pop_view_el);
     if (this.views.has(llvId)) return;
     const result = await this.pop.create({
       id: llvId,
@@ -197,6 +193,7 @@ export class LLVEngine {
       url: window.location.href,
       urlParams: Object.fromEntries(new URLSearchParams(window.location.search)),
       assigns: pop_view_el.getAttribute("data-pop-assigns")!,
+      mirrorId: mirrorId
     });
     // A rejected call resolves with ok: false (it does not throw). Bail on
     // failed or duplicate creates — wiring a fake view without a live
@@ -207,7 +204,6 @@ export class LLVEngine {
     }
     const liveEl = document.getElementById(llvId);
     if (liveEl?.matches("[data-pop-view]")) {
-      this.maybeSetupMirrorChannel(liveEl);
       setupFakeView(this.socket, this.views, this.popcornLink.socket, liveEl);
     } else {
       this.pop.destroy(llvId);
@@ -322,14 +318,15 @@ export class LLVEngine {
   private maybeSetupMirrorChannel(el: HTMLElement): void {
     const mirrorId = el.dataset.popMirrorId;
     if (!mirrorId || (mirrorId && this.channels[mirrorId])) return;
+    if (!el.dataset.popMirrorToken) return;
     const llvId = el.id;
+    console.log(el)
     const channel = this.llvMirrorSocket.channel(`llv:${mirrorId}`, {
       view: el.dataset.popView,
       token: el.dataset.popMirrorToken,
     });
     if (typeof mirrorId === "string") {
       this.channels[mirrorId] = channel;
-      this.pop.connectMirror(llvId, mirrorId);
     }
     channel
       .join()
--- a/local-live-view/assets/local_live_view/index.ts
+++ b/local-live-view/assets/local_live_view/index.ts
@@ -159,13 +159,13 @@ export class LLVEngine {
     // registerServerMessageListener / registerHooks comments).
     engine.registerServerMessageListener();
     registerNavigationHandlers(engine.socket, engine.views, engine.pop, engine.config);
+    engine.registerHooks();
     engine.bindFormsIfHostless();
     engine.connectPopcornSocket();
 
     await engine.bootPopcorn();
 
     engine.setupMirrorSync();
-    engine.registerHooks();
     engine.exposeGlobals();
     engine.patchOwner();
     registerCustomEventBindings(engine.socket);
--- a/local-live-view/assets/local_live_view/index.ts
+++ b/local-live-view/assets/local_live_view/index.ts
@@ -159,13 +159,13 @@ export class LLVEngine {
     // registerServerMessageListener / registerHooks comments).
     engine.registerServerMessageListener();
     registerNavigationHandlers(engine.socket, engine.views, engine.pop, engine.config);
-    engine.registerHooks();
     engine.bindFormsIfHostless();
     engine.connectPopcornSocket();
 
     await engine.bootPopcorn();
 
     engine.setupMirrorSync();
+    engine.registerHooks();
     engine.exposeGlobals();
     engine.patchOwner();
     registerCustomEventBindings(engine.socket);
@@ -249,9 +249,11 @@ export class LLVEngine {
     const maybeSetupMirrorChannel = (el: HTMLElement) => this.maybeSetupMirrorChannel(el);
     this.socket.hooks.LocalLiveView = {
       mounted() {
-        maybeSetupMirrorChannel(this.el);
         this.llvLastAssigns = this.el.getAttribute("data-pop-assigns")!;
-        if (pop.ready) mountView(this.el);
+        if (pop.ready) {
+          maybeSetupMirrorChannel(this.el);
+          mountView(this.el);
+        }
       },
       updated() {
         maybeSetupMirrorChannel(this.el);
--- a/local-live-view/assets/local_live_view/index.ts
+++ b/local-live-view/assets/local_live_view/index.ts
@@ -325,9 +325,7 @@ export class LLVEngine {
       view: popView,
       token: popMirrorToken,
     });
-    if (typeof mirrorId === "string") {
-      this.channels[mirrorId] = channel;
-    }
+    this.channels[mirrorId] = channel;
     channel
       .join()
       .receive("ok", () => {
--- a/local-live-view/assets/local_live_view/index.ts
+++ b/local-live-view/assets/local_live_view/index.ts
@@ -77,6 +77,14 @@ class PopcornClient {
     this.fire("reconnect sync", { action: "reconnected", id, payload: {} });
   }
 
+  connectMirror(id: string, mirror_id: string): void {
+    this.fire("connect mirror", {
+      action: "connect_mirror",
+      id,
+      payload: { mirror_id: mirror_id },
+    });
+  }
+
   updateAssigns(id: string, assigns: string): void {
     this.fire("update assigns", { action: "update_assigns", id, assigns });
   }
@@ -119,10 +127,20 @@ export class LLVEngine {
   // by __llvPushServer.
   private eventBusHooks = new Map<string, EventBusHook>();
   private popcornLink!: PopcornLink;
+  private llvMirrorSocket: PhoenixSocket;
 
   private constructor(socket: LLVSocket, config: LLVConfig) {
     this.socket = socket;
     this.config = config;
+
+    const Socket = this.socketClass();
+    const csrfToken = document.querySelector("meta[name='csrf-token']")?.getAttribute("content");
+
+    const llvSocket = new Socket("/llv_socket", {
+      params: { _csrf_token: csrfToken },
+    });
+    llvSocket.connect();
+    this.llvMirrorSocket = llvSocket;
   }
 
   /**
@@ -147,7 +165,7 @@ export class LLVEngine {
 
     await engine.bootPopcorn();
 
-    engine.setupMirrorChannels();
+    engine.setupMirrorSync();
     engine.exposeGlobals();
     engine.patchOwner();
     registerCustomEventBindings(engine.socket);
@@ -228,12 +246,15 @@ export class LLVEngine {
     const pop = this.pop;
     const mountView = (el: HTMLElement) => this.mountView(el);
     const unmountView = (el: HTMLElement) => this.unmountView(el);
+    const maybeSetupMirrorChannel = (el: HTMLElement) => this.maybeSetupMirrorChannel(el);
     this.socket.hooks.LocalLiveView = {
       mounted() {
+        maybeSetupMirrorChannel(this.el);
         this.llvLastAssigns = this.el.getAttribute("data-pop-assigns")!;
         if (pop.ready) mountView(this.el);
       },
       updated() {
+        maybeSetupMirrorChannel(this.el);
         const assigns = this.el.getAttribute("data-pop-assigns")!;
         if (assigns === this.llvLastAssigns) return;
         this.llvLastAssigns = assigns;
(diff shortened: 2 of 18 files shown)


local-live-view/lib/server/component.ex
```diff
@@ -42,62 +42,167 @@ defmodule LocalLiveView.Component do
...
 
-    unless is_binary(view) do
-      raise ArgumentError, """
-      <.local_live_view> expects view="..." parameter to be a string, got:
-        #{inspect(view)}
+    assigns = assign(assigns, id: id)
+
+    if mirror_exists?(view) do
+      ~H"""
+      <.live_component module={__MODULE__.Live} {assigns} />
```
mat-hek: Is the access to the socket the reason for introducing the live component? If so, can't we change the API so that instead of `endpoint={@endpoint}` the user would pass `socket={@socket}`?

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +42,167 @@ defmodule LocalLiveView.Component do
...
       phx-hook="LocalLiveView"
-      data-pop-mirror-token={@mirror_token}
       data-pop-assigns={encode_assigns(@comp_assigns)}
       phx-update="ignore"
     >
-    </div>
-    <%!-- Stub for sending events from client to server. See LLVEngine class. --%>
-    <div id={"#{@id}-llv-event-bus"} data-llv-event-bus-for={@id} phx-hook="LocalLiveViewEventBus" hidden>
+      <div id={"#{@id}-llv-event-bus"} data-llv-event-bus-for={@id} phx-hook="LocalLiveViewEventBus" hidden>
+      </div>
```
mat-hek: The even't bus can't be inside `data-pop-view`, otherwise it's removed once LLV renders into `data-pop-view`. I'm surprised it works now, but it's rather due to some favourable coincidence

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +42,167 @@ defmodule LocalLiveView.Component do
...
+  end
 
-    if mirror_exists?(view) && !assigns[:endpoint] do
-      raise ArgumentError, """
-      <.local_live_view> with mirror mechanism requires endpoint="..." parameter.
-      View "#{view}" uses mirror, but no endpoint was provided.
-      Add endpoint to your component call:
-        <.local_live_view view="#{view}" endpoint={@endpoint} />
-      """
+  defp render_dead(assigns) do
```
mat-hek: This name suggests we render inside a dead view, which is not true

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +42,167 @@ defmodule LocalLiveView.Component do
...
+        data-pop-mirror-id={@mirror_id}
+        data-pop-assigns={encode_assigns(@comp_assigns)}
+        phx-update="ignore"
+      >
+      </div>
+      <%!-- Stub for sending events from client to server. See LLVEngine class. --%>
+      <div id={"#{@id}-llv-event-bus"} data-llv-event-bus-for={@id} phx-hook="LocalLiveViewEventBus" hidden>
+      </div>
+      </div>
+      """
```
mat-hek: There's some code duplication between this and `render_dead`

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +42,167 @@ defmodule LocalLiveView.Component do
...
+    end
+
+    defp encode_assigns(assigns) when assigns == %{}, do: nil
+    defp encode_assigns(assigns), do: Jason.encode!(assigns)
```
mat-hek: We don't encode assigns with JSON anymore. Good example why to avoid code duplication :P

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +42,167 @@ defmodule LocalLiveView.Component do
...
+
+  ## Examples
+
+      # Explicit ID passed to component:
+      mirror_id = LocalLiveView.Component.mirror_id(socket, "my-custom-cart")
+
+      # Default ID derived from view name:
+      mirror_id = LocalLiveView.Component.mirror_id(socket, "Cart")
+  """
+  def mirror_id(%Phoenix.LiveView.Socket{id: socket_id}, view_or_id) do
```
mat-hek: this shouldn't be a public/documented function

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +42,167 @@ defmodule LocalLiveView.Component do
...
+      end
+
+    socket_id <> dom_id
+  end
+
+  defp default_id(name) when is_binary(name) do
+    "llv-" <> String.replace(name, ~r/[^A-Za-z0-9_-]/, "-")
+  end
+
+  defp default_id(_), do: "llv-view"
```
mat-hek: Don't we always get a string here? Why is this clause needed?

mat-hek (changes requested):
I only looked at `component.ex` for now

local-live-view/lib/server/component.ex
FranekKubis: that was a leftover from earlier implementation

local-live-view/lib/server/component.ex
FranekKubis: I made it common now

local-live-view/lib/server/component.ex
FranekKubis: This can be used outside of the module when someone has access to socket and the ids of its child LLVs.
It is better to assemble mirror_id using API function rather than having to know how the id is created.

local-live-view/lib/server/component.ex
FranekKubis: It is indeed not needed.

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,153 @@ defmodule LocalLiveView.Component do
...
     >
     </div>
     <%!-- Stub for sending events from client to server. See LLVEngine class. --%>
     <div id={"#{@id}-llv-event-bus"} data-llv-event-bus-for={@id} phx-hook="LocalLiveViewEventBus" hidden>
     </div>
+    </div>
     """
   end
 
+  @doc false
```
mat-hek, line 157:
```suggestion
```

local-live-view/lib/server/component.ex
mat-hek: ok, let's add a typespec then

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,153 @@ defmodule LocalLiveView.Component do
...
-    comp_assigns = Map.drop(assigns, [:__changed__, :view])
+  @doc false
+  def render_markup(assigns) do
+    validate_assigns!(assigns)
 
-    mirror_token =
-      if mirror_exists?(view),
-        do: LocalLiveView.MirrorToken.sign(assigns.endpoint, view, assigns.id)
+    comp_assigns =
+      Map.drop(assigns, [:__changed__, :view, :flash, :id, :mirror_id, :mirror_token])
```
mat-hek: Should we drop id? In the docs we say we forward it

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,153 @@ defmodule LocalLiveView.Component do
...
+          nil
+        end
+
+      socket =
+        assign(socket,
+          view: view,
+          id: assigns.id,
+          mirror_token: mirror_token,
+          mirror_id: mirror_id
+        )
```
mat-hek, line 107: we never assign other assigns, so `@comp_assigns` in render_markup are empty

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,153 @@ defmodule LocalLiveView.Component do
...
+      assign(assigns,
+        mirror_token: nil,
+        mirror_id: nil
+      )
 
-    if mirror_exists?(view) && !assigns[:endpoint] do
+    render_markup(assigns)
+  end
+
+  defmodule Live do
```
mat-hek:
```suggestion
  defmodule Mirrored do
```

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,153 @@ defmodule LocalLiveView.Component do
...
+
+  def mirror_id(socket_id, view_or_id) when is_binary(socket_id) and is_binary(view_or_id) do
+    dom_id =
+      if String.starts_with?(view_or_id, "llv-") do
+        view_or_id
+      else
+        default_id(view_or_id)
+      end
+
+    socket_id <> dom_id
```
mat-hek: maybe separate them?


```suggestion
    socket_id <> "-" <> dom_id
```

mat-hek (changes requested):

local-live-view/assets/local_live_view/index.ts
```diff
@@ -228,12 +246,15 @@ export class LLVEngine {
     const pop = this.pop;
     const mountView = (el: HTMLElement) => this.mountView(el);
     const unmountView = (el: HTMLElement) => this.unmountView(el);
+    const maybeSetupMirrorChannel = (el: HTMLElement) => this.maybeSetupMirrorChannel(el);
     this.socket.hooks.LocalLiveView = {
       mounted() {
+        maybeSetupMirrorChannel(this.el);
```
mat-hek: This calls `connectMirror` under the hood, which sends action to popcorn, but popcorn may be not ready yet (see 2 lines below). Maybe it should be a part of `mountView`?

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,153 @@ defmodule LocalLiveView.Component do
...
+      # Default ID derived from view name:
+      mirror_id = LocalLiveView.Component.mirror_id(socket, "Cart")
+  """
+  def mirror_id(%Phoenix.LiveView.Socket{id: socket_id}, view_or_id) do
+    mirror_id(socket_id, view_or_id)
+  end
+
+  def mirror_id(socket_id, view_or_id) when is_binary(socket_id) and is_binary(view_or_id) do
+    dom_id =
+      if String.starts_with?(view_or_id, "llv-") do
```
mat-hek: Does this work when someone passes a custom id to LLV? We don't require the id to start with `llv-`

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,153 @@ defmodule LocalLiveView.Component do
...
 
-    unless is_binary(view) do
-      raise ArgumentError, """
-      <.local_live_view> expects view="..." parameter to be a string, got:
-        #{inspect(view)}
+    assigns = assign(assigns, id: id)
+
+    if mirror_exists?(view) do
+      ~H"""
+      <.live_component module={__MODULE__.Live} {assigns} />
```
mat-hek: This will break when someone tries to render mirrored LLV outside of a LiveView, as live components cannot be rendered in regular views AFAIK

mat-hek (changes requested):
Also, Claude told me that this broke burrito example. Haven't checked this, did you? BTW, it would be worth running all examples' CIs when LLV changes, as these are our main tests.

local-live-view/lib/server/component.ex
FranekKubis: We were dropping it before, so I assumed the id is not needed in comp_assigns.
Also we forward it as a DOM element id.

local-live-view/lib/server/component.ex
FranekKubis: I am aware of that - I planned to leave it as a known limitation.

local-live-view/lib/server/component.ex
mat-hek: We're not dropping it on main: https://github.com/software-mansion/popcorn/blob/main/local-live-view/lib/server/component.ex#L68 Even if it's a DOM element id, it's not accessible in assigns, and we claim so here: https://github.com/software-mansion/popcorn/blob/fk/llv-id-generation/local-live-view/lib/server/component.ex#L24 . I guess we should pass it to assigns, as that's what live component does

local-live-view/lib/server/component.ex
mat-hek: If so, we should raise a clear error in that case. Would the limitation apply only to mirrored live views?

local-live-view/assets/local_live_view/index.ts
FranekKubis: I moved it to if pop.ready scope.

local-live-view/lib/server/component.ex
FranekKubis: Yes - I removed that part.

local-live-view/lib/server/component.ex
FranekKubis: Yes

local-live-view/assets/local_live_view/index.ts
mat-hek: I don't think it will work - if popcorn is not ready in time, it never runs. `mountView` is different, because we run 'startup scan' once popcorn is ready: https://github.com/software-mansion/popcorn/blob/main/local-live-view/assets/local_live_view/index.ts#L390.

FranekKubis: So moved it to mountView function

local-live-view/lib/server/component.ex
FranekKubis: fixed

local-live-view/assets/local_live_view/index.ts
```diff
@@ -228,12 +247,14 @@ export class LLVEngine {
...
     const mountView = (el: HTMLElement) => this.mountView(el);
     const unmountView = (el: HTMLElement) => this.unmountView(el);
+    const maybeSetupMirrorChannel = (el: HTMLElement) => this.maybeSetupMirrorChannel(el);
     this.socket.hooks.LocalLiveView = {
       mounted() {
         this.llvLastAssigns = this.el.getAttribute("data-pop-assigns")!;
         if (pop.ready) mountView(this.el);
       },
       updated() {
+        maybeSetupMirrorChannel(this.el);
```
mat-hek: Why do we setup the channel here? Isn't it enough to setup it on mount? If so, can we remove the `connect_mirror` event and just put the id in `pop.create`?

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,149 @@ defmodule LocalLiveView.Component do
...
+  * a view module name / string (calculates the default DOM ID)
+  * an explicit custom DOM ID provided to `<.local_live_view id="..." />`
+
+  ## Examples
+
+      # Explicit ID passed to component:
+      mirror_id = LocalLiveView.Component.mirror_id(socket, "my-custom-cart")
+
+      # Default ID derived from view name:
+      mirror_id = LocalLiveView.Component.mirror_id(socket, "Cart")
```
mat-hek: This no longer works, does it?

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,149 @@ defmodule LocalLiveView.Component do
...
+  """
+  @spec mirror_id(%Phoenix.LiveView.Socket{id: String.t()}, String.t()) :: String.t()
+  def mirror_id(%Phoenix.LiveView.Socket{id: socket_id}, view_or_id) do
```
mat-hek: as per above

```suggestion
  def mirror_id(%Phoenix.LiveView.Socket{id: socket_id}, id) do
```

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,149 @@ defmodule LocalLiveView.Component do
...
+
+  ## Examples
+
+      # Explicit ID passed to component:
+      mirror_id = LocalLiveView.Component.mirror_id(socket, "my-custom-cart")
+
+      # Default ID derived from view name:
+      mirror_id = LocalLiveView.Component.mirror_id(socket, "Cart")
+  """
+  @spec mirror_id(%Phoenix.LiveView.Socket{id: String.t()}, String.t()) :: String.t()
```
mat-hek:
```suggestion
  @spec mirror_id(Phoenix.LiveView.Socket.t(), String.t()) :: String.t()
```

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,149 @@ defmodule LocalLiveView.Component do
...
+      view = assigns[:view]
+      mirror_id = LLVComponent.mirror_id(socket, assigns.id)
+
+      mirror_token =
+        if Phoenix.LiveView.connected?(socket) do
+          endpoint = socket.endpoint || LLVComponent.resolve_default_endpoint()
+          LocalLiveView.MirrorToken.sign(endpoint, view, mirror_id)
+        else
+          nil
+        end
```
mat-hek: I don't see a reason to sign the ID on each update, as it never changes. I think we can do

```suggestion
        cond do
          token = socket.assigns[:mirror_token] ->
            token
          Phoenix.LiveView.connected?(socket) ->
            endpoint = socket.endpoint || LLVComponent.resolve_default_endpoint()
            LocalLiveView.MirrorToken.sign(endpoint, view, mirror_id)
          true ->
            nil
        end
```

mat-hek (changes requested):

local-live-view/lib/server/component.ex
mat-hek: Or even better generate the ID only if it's not generated yet

local-live-view/assets/local_live_view/index.ts
FranekKubis: I removed the connectMirror event and move the mirror connection functionality to mount.

local-live-view/assets/local_live_view/index.ts
```diff
@@ -322,14 +318,14 @@ export class LLVEngine {
   private maybeSetupMirrorChannel(el: HTMLElement): void {
```
mat-hek: nit: this could accept llvId, popView and popMirrorToken, so that the extraction stays in mountView

local-live-view/assets/local_live_view/index.ts
mat-hek: So the call to `maybeSetupMirrorChannel` here is redundant, isn't it?

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,159 @@ defmodule LocalLiveView.Component do
...
+  def resolve_default_endpoint do
+    Application.get_env(:local_live_view, :default_endpoint)
+  end
+
+  @doc """
+  Assembles the deterministic `mirror_id` for a given LocalLiveView component
+  within a parent LiveView process.
+
+  Accepts the parent `socket` (or directly its `socket.id`), and either:
+  * a view module name / string (calculates the default DOM ID)
```
mat-hek: to update

local-live-view/assets/local_live_view/index.ts
```diff
@@ -289,46 +306,37 @@ export class LLVEngine {
...
       if (channel) {
         channel.push(eventName, payload);
       }
     };
   }
 
+  private maybeSetupMirrorChannel(el: HTMLElement): void {
+    const mirrorId = el.dataset.popMirrorId;
+    if (!mirrorId || (mirrorId && this.channels[mirrorId])) return;
+    if (!el.dataset.popMirrorToken) return;
```
mat-hek: nit

```suggestion
    if (!mirrorId || this.channels[mirrorId] || !el.dataset.popMirrorToken) return;
```

local-live-view/lib/server/component.ex
```diff
@@ -42,62 +41,159 @@ defmodule LocalLiveView.Component do
...
 
-    unless is_binary(view) do
-      raise ArgumentError, """
-      <.local_live_view> expects view="..." parameter to be a string, got:
-        #{inspect(view)}
+    assigns = assign(assigns, id: id)
+
+    if mirror_exists?(view) do
+      ~H"""
+      <.live_component module={__MODULE__.Mirrored} {assigns} />
```
mat-hek, line 50: We still don't throw a readable error when not in LV

examples/local-lv-burrito/lib/burrito_web/live/sync_table_live.ex
```diff
@@ -16,12 +16,20 @@ defmodule BurritoWeb.Live.SyncTableLive do
...
+        true ->
+          mirror_id =
+            LocalLiveView.Component.mirror_id(llv_rendering_socket_id, "llv-BurritoLive")
+
+          Phoenix.PubSub.subscribe(Burrito.PubSub, "llv_mirror:BurritoLive:#{mirror_id}")
+          LocalLiveView.Channel.get_mirror_assigns(mirror_id)
+
+        false ->
+          %{}
+      end
```
mat-hek: why not if? :P

local-live-view/assets/local_live_view/index.ts
```diff
@@ -119,10 +121,20 @@ export class LLVEngine {
...
     this.config = config;
+
+    const Socket = this.socketClass();
+    const csrfToken = document.querySelector("meta[name='csrf-token']")?.getAttribute("content");
+
+    const llvSocket = new Socket("/llv_socket", {
+      params: { _csrf_token: csrfToken },
+    });
+    llvSocket.connect();
+    this.llvMirrorSocket = llvSocket;
```
mat-hek: We now always start the socket, no matter if there are mirrors or not. I think we can lazily start it in `maybeSetupMirrorChannel`

mat-hek (commented):
Looks good, minor tweaks left. Note that there were quite some bugs, not covered by the tests. For now I'd at least run examples' CIs when LLV code changes.

local-live-view/lib/server/component.ex
FranekKubis: I am inable to catch this exception on this level because the error here will be raised by the page controller mechanism inside phoenix.
Right now it is failing like this:
```
 1) test GET / renders the LocalLiveView mount points (LocalThermostatWeb.PageControllerTest)
     test/local_thermostat_web/controllers/page_controller_test.exs:4
     ** (ArgumentError) cannot convert component LocalLiveView.Component.Mirrored with id "llv-ClockLive" to HTML.

     A component must always be returned directly as part of a LiveView template.

     For example, this is not allowed:

         <%= content_tag :div do %>
           <.live_component module={SomeComponent} id="myid" />
         <% end %>

     That's because the component is inside `content_tag`. However, this works:

         <div>
           <.live_component module={SomeComponent} id="myid" />
         </div>

     Components are also allowed inside Elixir's special forms, such as
     `if`, `for`, `case`, and friends.

         <%= for item <- items do %>
           <.live_component module={SomeComponent} id={item} />
         <% end %>

     However, using other module functions such as `Enum`, will not work:

         <%= Enum.map(items, fn item -> %>
           <.live_component module={SomeComponent} id={item} />
         <% end %>
     
     code: conn = get(conn, ~p"/")
     stacktrace:
       (phoenix_live_view 1.1.28) lib/phoenix_live_view/engine.ex:20: Phoenix.HTML.Safe.Phoenix.LiveView.Component.to_iodata/1
       (phoenix_live_view 1.1.28) lib/phoenix_live_view/engine.ex:146: Phoenix.HTML.Safe.Phoenix.LiveView.Rendered.recur_iodata/3
       (phoenix 1.8.5) lib/phoenix/controller.ex:1015: anonymous fn/5 in Phoenix.Controller.template_render_to_iodata/4
       (telemetry 1.4.1) /Users/franek/Desktop/Repositories/popcorn/examples/local-lv-thermostat/deps/telemetry/src/telemetry.erl:359: :telemetry.span/3
       (phoenix 1.8.5) lib/phoenix/controller.ex:981: Phoenix.Controller.render_and_send/4
       (local_thermostat 0.1.0) lib/local_thermostat_web/controllers/page_controller.ex:1: LocalThermostatWeb.PageController.action/2
       (local_thermostat 0.1.0) lib/local_thermostat_web/controllers/page_controller.ex:1: LocalThermostatWeb.PageController.phoenix_controller_pipeline/2
       (phoenix 1.8.5) lib/phoenix/router.ex:416: Phoenix.Router.__call__/5
       (local_thermostat 0.1.0) lib/local_thermostat_web/endpoint.ex:1: LocalThermostatWeb.Endpoint.plug_builder_call/2
       (local_thermostat 0.1.0) lib/local_thermostat_web/endpoint.ex:1: LocalThermostatWeb.Endpoint.call/2
       (phoenix 1.8.5) lib/phoenix/test/conn_test.ex:225: Phoenix.ConnTest.dispatch/5
       test/local_thermostat_web/controllers/page_controller_test.exs:5: (test)

```

examples/local-lv-burrito/lib/burrito_web/live/sync_table_live.ex
FranekKubis: it can be `if`

local-live-view/assets/local_live_view/index.ts
```diff
@@ -289,44 +304,50 @@ export class LLVEngine {
...
+  }: MirrorChannelOptions): void {
+    if (!mirrorId || this.channels[mirrorId] || !popMirrorToken) return;
+    const socket = this.llvMirrorSocket ?? this.setupMirrorSocket();
+    const channel = socket.channel(`llv:${mirrorId}`, {
+      view: popView,
+      token: popMirrorToken,
+    });
+    if (typeof mirrorId === "string") {
+      this.channels[mirrorId] = channel;
+    }
```
mat-hek: seems always true

```suggestion
    this.channels[mirrorId] = channel;
```

mat-hek (approved):
🎉

PR #713, opened 2026-07-29
[LLV] Enable treeshake for kanban, small treeshake fixes

mat-hek:


popcorn-2/elixir/lib/popcorn.ex
```diff
@@ -89,8 +89,15 @@ defmodule Popcorn do
...
-        %{behaviour_impls: LocalLiveView}
+        # Phoenix resolves Jason via `Phoenix.json_library/0`, what makes it
+        # invisible for treeshaking. Jason is low overhead and needed for
+        # most Popcorn stuff anyway, so at least for now we keep it unconditionally.
+        Jason,
+        # LiveView/LLV modules are resolved via Heex templates, which are just strings,
+        # what makes them invisible for treeshaking. For now we keep all LLV impls.
+        %{behaviour_impls: LocalLiveView},
+        # Popcorn.Init is called by the boot module, which is ignored (see `:ignore`)
+        Popcorn.Init
```
jgonet, line 100: Shouldn't we keep those things out of library and in the local-live-view?

mat-hek: Some of them are very popcorn-specific (Popcorn.Init), so they must be there. Jason and LLV behaviours could sit in LLV if we exposed an API for this configuration, but having that in many places is way harder to follow. Also, LLV is the only relevant use case for tree shaking for now, so I decided not to extract it.

jgonet (approved):

PR #716, opened 2026-07-29
treeshake: prepare for otp

jgonet:
- avoids adding helper when it's not needed.
- sidesteps the problem of otp packager not handling builtin apps by providing behaviour information from host (this has its drawbacks, maybe we should redesign packager in the future).

### DIFF
--- a/popcorn/elixir/lib/treeshake/call_graph.ex
+++ b/popcorn/elixir/lib/treeshake/call_graph.ex
@@ -150,18 +150,25 @@ defmodule Treeshake.CallGraph do
       Enum.map(info.behaviour_impls, &{info.module, &1})
     end)
     |> Enum.flat_map(fn {module, behaviour} ->
-      case module_index[behaviour] do
-        %{abstraction: {:behaviour, callbacks}} ->
-          callbacks
-
-        nil ->
-          []
-
-        _non_behaviour ->
-          raise "Module #{module} implements #{behaviour} as if it was a behaviour, but it's not"
-      end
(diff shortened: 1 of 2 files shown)


popcorn/elixir/lib/treeshake/call_graph.ex
```diff
@@ -150,20 +150,27 @@ defmodule Treeshake.CallGraph do
...
+      behaviour_callbacks(module_index[behaviour], module, behaviour)
       |> Enum.map(fn {f, a} -> {module, f, a} end)
     end)
   end
 
+  defp behaviour_callbacks(%{abstraction: {:behaviour, callbacks}}, _module, _behaviour),
+    do: callbacks
+
+  defp behaviour_callbacks(nil, _module, behaviour) do
+    with {:module, ^behaviour} <- Code.ensure_loaded(behaviour),
```
mat-hek, line 162: Don't we scan all the beam files that can be loaded? If not, we probably should

popcorn/elixir/lib/treeshake/treeshake.ex
```diff
@@ -217,8 +217,11 @@ defmodule Treeshake do
     end
 
     stats = Treeshake.Shaker.shake(opts, call_graph, module_index)
+    stub_functions = Map.get(opts, :stub_removed_functions, false)
+    stub_modules = Map.get(opts, :stub_removed_modules, false)
```
mat-hek, line 221: Now we set the default here and in the shaker. Let's do that once, in parse_opts

PR #717, opened 2026-07-29
otp: treeshake packaged applications

jgonet:
It's a bit silly, we package the treeshaker in app so TS plugin can compile
and run it. The better way would be to just create treeshaker app or combine files
into a script.

We could also move the packaging part into Elixir but frankly, it already needs
assets from JS.


mat-hek:
Don't know what is this but sounds very weird :P

PR #721, opened 2026-08-01
[LLV] Fix phx-hook mounting inside LocalLiveView

edun17:
Fake views omit data-phx-session, so stock ownsElement never claims the subtree and addHook skips child hooks. Own nodes by container containment.


local-live-view/assets/local_live_view/types.ts
```diff
@@ -68,6 +68,7 @@ export interface LLVView {
     opts: object,
   ): void;
   addHook: (el: Element) => unknown;
+  ownsElement?: (el: Element) => boolean;
```
FranekKubis: I don't think we need that.

local-live-view/priv/static/local_live_view.js
```diff
@@ -524,6 +524,12 @@ function setupFakeView(socket, views, popcornSocket, pop_view_el) {
     // diffs, out-of-band "diff" frames, ref bookkeeping — runs through the
     // stock Channel/Push machinery over the PopcornTransport.
     view.channel = popcornSocket.channel(`lv:${llvId}`);
+    // LLV roots have no data-phx-session, so stock ownsElement (closestViewEl) would
+    // assign children to the host LiveView or reject them entirely — and addHook
+    // would skip every phx-hook under this container. Own the subtree by containment.
+    view.ownsElement = function (el) {
```
FranekKubis, line 530: This is good implementation, lets move it to `view_setup.ts` and recreate `priv/static/local_live_view.js` with `pnpm -F ./local-live-view build`.

local-live-view/assets/local_live_view/view_setup.ts
```diff
@@ -20,6 +20,13 @@ export function setupFakeView(
   // stock Channel/Push machinery over the PopcornTransport.
   view.channel = popcornSocket.channel(`lv:${llvId}`);
 
+  // LLV roots have no data-phx-session, so stock ownsElement (closestViewEl) would
+  // assign children to the host LiveView or reject them entirely — and addHook
+  // would skip every phx-hook under this container. Own the subtree by containment.
+  view.ownsElement = function (this: LLVView, el: Element) {
```
FranekKubis: Lets use the other implementation here:
```
view.ownsElement = function (el) {
        return this.el === el || this.el.contains(el);
    };
```

FranekKubis (changes requested):

FranekKubis (approved):

PR #726, opened 2026-08-05
[LLV] Cleaning before release

FranekKubis:
Removed generated files from repo. Now llv.build is checking if local_live_view is path or hex dependency and behaves accordingly.


local-live-view/mix.exs
```diff
@@ -51,6 +85,17 @@ defmodule LocalLiveView.MixProject do
     ]
   end
 
+  # In this repo we develop against the Popcorn source tree, so API changes on
+  # both sides land in one commit. A path dep cannot be published, so the
+  # release pipeline sets LLV_RELEASE=1 to depend on the published Popcorn.
+  defp popcorn_dep do
+    if System.get_env("LLV_RELEASE") do
```
mat-hek: This env is not going to be set when a user fetches the package. I think that the approach with switching the dep to path temporarily when needed and restoring it once popcorn is released is not that bad.

local-live-view/pages/getting-started/installation.md
```diff
@@ -16,7 +16,7 @@ Add `:local_live_view` to your `mix.exs`:
 defp deps do
   [
     # ...existing deps...
-    {:local_live_view, github: "software-mansion/popcorn", sparse: "local-live-view"}
+    {:local_live_view, , "~> 0.1"}
```
mat-hek: `~> 0.1` would automatically bump someone's LLV to 0.2 when it's released, and it's likely to have breaking changes

```suggestion
    {:local_live_view, "~> 0.1.0"}
```

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -501,17 +507,26 @@ defmodule Mix.Tasks.Llv.Install do
...
   end
 
+  # The generated local/ project depends on LocalLiveView as well (it compiles
+  # the WASM side), so it has to declare the dep the same way the host app got
+  # it: a relative path for a path dep, a version requirement otherwise.
+  defp llv_dep_from_local do
+    if llv_path_dep?() do
+      ~s|{:local_live_view, path: "#{llv_path_from_local()}"}|
+    else
+      ~s|{:local_live_view, "~> #{@llv_version_requirement}"}|
```
mat-hek: If we put a constraint on LLV in the local mix.exs, the user has to bump it in two places (the host mix.exs and the local mix.exs)

```suggestion
      ~s|{:local_live_view, ">= 0.0.0"}|
```

local-live-view/mix.exs
```diff
@@ -1,23 +1,57 @@
...
   def project do
     [
       app: :local_live_view,
-      version: "0.1.0",
+      version: @version,
       elixir: "~> 1.17",
       elixirc_paths: elixirc_paths(Mix.target()),
       start_permanent: Mix.env() == :prod,
       aliases: aliases(),
       preferred_cli_target: [docs: :docs],
```
mat-hek: Just realized that we'll also need

```suggestion
      preferred_cli_target: [
        docs: :docs
        "hex.publish": :docs
      ],
```

Maybe it would be better to call the docs target `:all`?

examples/local-lv-kanban/mix.exs
```diff
@@ -84,7 +84,14 @@ defmodule LocalLvKanban.MixProject do
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       test: ["llv.build", "ecto.create --quiet", "ecto.migrate --quiet", "test"],
       "assets.setup": ["tailwind.install --if-missing", "esbuild.install --if-missing"],
-      "assets.build": ["compile", "tailwind local_lv_kanban", "esbuild local_lv_kanban"],
+      # llv.assets first: esbuild resolves "local_live_view" to the bundle in
+      # deps/, which for an in-repo path dep is generated rather than committed.
+      "assets.build": [
+        "llv.assets",
```
mat-hek: Why does kanban have it and other deps don't? It'd be great to avoid adding another llv-specific command here, maybe we can move it to `llv.build`? Also, claude claims that assets.deploy doesn't run assets.build so we'd need to add it there as well

local-live-view/lib/mix/tasks/llv.assets.ex
```diff
@@ -0,0 +1,167 @@
...
+  end
+
+  defp missing(static_dir) do
+    Enum.reject(@artifacts, &File.regular?(Path.join(static_dir, &1)))
+  end
+
+  defp sources_hash(llv_dir) do
+    sources =
+      Path.wildcard(Path.join(llv_dir, "assets/local_live_view/**/*.ts")) ++
+        Enum.map(@extra_sources, &Path.join(llv_dir, &1))
```
mat-hek: We don't hash `AtomVM.mjs` and `AtomVM.wasm`, can't it make `fresh?` return `true` even if they changed?

local-live-view/lib/mix/tasks/llv.assets.ex
```diff
@@ -0,0 +1,167 @@
+defmodule Mix.Tasks.Llv.Assets do
+  use Mix.Task
+
+  @shortdoc "Ensures LocalLiveView's prebuilt JS assets are available"
+
+  @moduledoc """
```
mat-hek: Shouldn't be public. BTW, we should have other tasks in the docs

mat-hek (changes requested):
Generally, I'm not sure if we should change the way the assets are shipped just before the release. These things are tricky and hard to verify they work for everyone, and since it's the first release we're likely to have other issues too. We can as well do that after the release.
Another thing, Claude found quite some more issues apart from my comments, and they sound real. If you didn't tell Claude to review, please do so or have a look at what it reported for me:

- Sparse git deps can no longer build — [llv.assets.ex:87](vscode-webview://1g7lpu1cnva7aj88cau3lrdr7m1a2f58lrjcvcsjrdad4k88f38j/local-live-view/lib/mix/tasks/llv.assets.ex#L87)
On main, the docs recommended {:local_live_view, github: "software-mansion/popcorn", sparse: "local-live-view"}, and it worked with zero Node toolchain because priv/static/ shipped in git.
On this branch priv/static/ is gitignored, so that same dep arrives without the bundle; llv.assets sees an assets/ directory and takes the build-from-source path (build!), which starts with pnpm install.
Here's the trap: [local-live-view/package.json](vscode-webview://1g7lpu1cnva7aj88cau3lrdr7m1a2f58lrjcvcsjrdad4k88f38j/local-live-view/package.json) declares "eslint": "catalog:" and "typescript": "catalog:". The catalog: specifier is resolved from pnpm-workspace.yaml — which lives at the monorepo root, outside the sparse checkout. Mix's sparse fetch only materializes the local-live-view/ subtree.
So pnpm install fails with ERR_PNPM_CATALOG_ENTRY_NOT_FOUND, and mix llv.build aborts with no working fallback.
This contradicts the task's own moduledoc, which claims support for "source checkout (path dep, git dep, this repo)". (One sub-claim was refuted during verification: the emsdk/AtomVM concern does not apply outside the monorepo, because @swmansion/popcorn resolves from npm there and its published tarball ships a prebuilt dist/AtomVM.wasm.) Fix: replace the catalog: entries with concrete versions, or drop the git-dep claim and detect/reject sparse checkouts with a clear message.
- Manual releases build from the wrong ref — [release_llv.yml:56](vscode-webview://1g7lpu1cnva7aj88cau3lrdr7m1a2f58lrjcvcsjrdad4k88f38j/.github/workflows/release_llv.yml#L56)
The workflow has two triggers: pushing an llv-v* tag, and workflow_dispatch with a tag input (default dry_run: true).
The Docker build passes --build-arg GIT_REF=${{ github.ref_name }}, and docker/release.Dockerfile does a fresh git clone --branch "$GIT_REF" from GitHub — the image ignores the workflow's own checkout entirely.
On a tag push, ref_name is the tag — correct. On workflow_dispatch, ref_name is the branch you dispatched from, not the tag input.
So dispatching with tag=llv-v0.1.0, dry_run=false passes the version check (the branch's declared version matches the input), then clones and publishes whatever the branch tip is at that moment — possibly commits pushed after you clicked, or a branch where the tag doesn't exist — to Hex under that version number.
Fix is one expression: GIT_REF=${{ inputs.tag || github.ref_name }} (git clone --branch accepts tag names). Note release_hex.yml has the identical latent bug.
- pnpm install is skipped whenever node_modules exists — [llv.assets.ex:101](vscode-webview://1g7lpu1cnva7aj88cau3lrdr7m1a2f58lrjcvcsjrdad4k88f38j/local-live-view/lib/mix/tasks/llv.assets.ex#L101)
build!/2 runs pnpm install only unless File.dir?(node_modules) — presence of the directory is treated as "dependencies are correct".
But package.json is hashed by sources_hash, so bumping a JS dep version changes the hash and correctly triggers a rebuild…
…which then runs pnpm run build against the old node_modules, because the install step was skipped. No error, wrong dependency versions in the output bundle.
It also drifts from the sibling entry point: the build alias in [local-live-view/mix.exs](vscode-webview://1g7lpu1cnva7aj88cau3lrdr7m1a2f58lrjcvcsjrdad4k88f38j/local-live-view/mix.exs) uses a pnpm_install/1 helper that installs unconditionally. Two entry points, two staleness behaviors.
Fix: run pnpm install unconditionally in build! — with a warm store and an up-to-date tree it's roughly a second, and it makes the "hash changed → rebuild" path actually sound.
- The release workflow is a third drifting copy — [release_llv.yml:33](vscode-webview://1g7lpu1cnva7aj88cau3lrdr7m1a2f58lrjcvcsjrdad4k88f38j/.github/workflows/release_llv.yml#L33)
The inline Python version-alignment check is a near-verbatim copy of the same step in release_hex.yml and release_npm.yml (only tag prefix and file paths differ), and the docker-build + publish steps mirror release_hex.yml almost line for line.
Copies drift. This branch already contains the proof: the MIX_ENV=dev mix hex.publish docs line was copied from release_hex.yml, where it works because popcorn's ex_doc has no targets: filter — and it breaks in LLV's context (finding 2). That's drift happening at copy time.
Going forward, any fix to the pre-release regex, dry-run handling, or the docs workaround applied in one workflow silently misses the other two, so an llv-v* release runs under different rules than a v* release.
Fix: factor the version check into a composite action under .github/actions/ (the repo already uses that directory) parameterized by tag prefix and file paths, and/or make the docker-publish sequence a reusable workflow_call workflow both tag namespaces invoke.

examples/local-lv-kanban/mix.exs
FranekKubis: the `llv.assets` here is not needed - i reverted the change.

local-live-view/mix.exs
FranekKubis: changed to: 
```
if File.dir?(@popcorn_source) and System.get_env("LLV_RELEASE") == nil do
  {:popcorn, path: @popcorn_source, targets: [:wasm, :all]}
else
  {:popcorn, "~> #{@popcorn_version}", targets: [:wasm, :all]}
end
```

FranekKubis:
Issues (from last big comment) fixed:
- Sparse git deps can no longer build — done
- pnpm install is skipped whenever node_modules exists - done

FranekKubis:
Last two things fixed:
- Manual releases build from the wrong ref - done
- The release workflow is a third drifting copy - done (added composite action)

Also: I'd rather keep the changes to generated artifacts as they are on this branch because it is closely related to releasing.

Also: Generating the code from llv as git sparse dependecy is working fine, ~~I am also still checking overall final state of this branch with claude - when its ok I will re-request for another review.~~ done

local-live-view/lib/mix/tasks/llv.assets.ex
FranekKubis: Added them to hash.

local-live-view/mix.exs
```diff
@@ -32,25 +69,35 @@ defmodule LocalLiveView.MixProject do
...
+  defp popcorn_dep do
+    if File.dir?(@popcorn_source) and System.get_env("LLV_RELEASE") == nil do
+      {:popcorn, path: @popcorn_source, targets: [:wasm, :all]}
+    else
+      {:popcorn, "~> #{@popcorn_version}", targets: [:wasm, :all]}
+    end
+  end
+
   defp docs do
     [
```
mat-hek, line 102: So that links to LLV source from hexdocs work. Assumes we're going to tag it `llv-0.1.0` on publish

```suggestion
    [
      source_ref: "llv-#{@version}"
```

local-live-view/mix.exs
```diff
@@ -1,23 +1,60 @@
 defmodule LocalLiveView.MixProject do
   use Mix.Project
 
+  @version "0.1.0"
```
mat-hek: Let's publish an RC first, check if it works and then publish 0.1.0

```suggestion
  @version "0.1.0-rc0"
```

local-live-view/mix.exs
```diff
@@ -1,23 +1,60 @@
...
+  @popcorn_version "0.3.3"
+  @github "https://github.com/software-mansion/popcorn"
```
mat-hek, line 6:
```suggestion
  @github "https://github.com/software-mansion/popcorn/local-live-view"
```

local-live-view/lib/mix/tasks/llv.assets.ex
```diff
@@ -0,0 +1,164 @@
+defmodule Mix.Tasks.Llv.Assets do
```
mat-hek, line 1:
```suggestion
defmodule Mix.Tasks.Llv.Assets do
  @moduledoc false
```

local-live-view/lib/mix/tasks/llv.build.ex
```diff
@@ -8,11 +8,15 @@ defmodule Mix.Tasks.Llv.Build do
 
   ## Steps
 
-    1. Copies Popcorn runtime files (`iframe.mjs`, `AtomVM.mjs`, `AtomVM.wasm`)
+    1. Makes sure `deps/local_live_view/priv/static/` holds the JS bundle. With
+       the published package there is nothing to do, since the bundle ships in
+       the tarball; a source checkout is bundled on the spot, which needs
+       Node.js and pnpm.
```
mat-hek: The user doesn't care about this

local-live-view/lib/mix/tasks/llv.install.ex
```diff
@@ -501,17 +456,50 @@ defmodule Mix.Tasks.Llv.Install do
...
+    Mix.Project.config()[:deps]
+    |> List.wrap()
+    |> Enum.find_value(fn
+      {:local_live_view, opts} when is_list(opts) -> {nil, opts}
+      {:local_live_view, req} when is_binary(req) -> {req, []}
+      {:local_live_view, req, opts} when is_binary(req) and is_list(opts) -> {req, opts}
+      _ -> nil
+    end)
+  end
+
```
mat-hek, line 484: This is too complex and still copies the hex requirement instead of setting ">= 0.0.0". What we should do:
- if there's a hex requirement, put ">= 0.0.0"
- if there's a git dep, copy it
- if there's a path dep, copy it and prepend `../`

I'd leave all other options untouched, as it may bring more mess than profit

mat-hek (changes requested):
Also, claude says that `local-live-view/priv/static/local_live_view.js` is still in the repo and should be removed

local-live-view/lib/mix/tasks/llv.build.ex
FranekKubis: removed this doc step

local-live-view/mix.exs
mat-hek: actually it should be 
```suggestion
    [
       source_ref: "llv-v#{@version}/local-live-view"
```

local-live-view/mix.exs
mat-hek: I just realized it won't work... Either set source_ref to `llv-v#{version}/local-live-view` or use [source_url_pattern](https://ex-doc.hexdocs.pm/ExDoc.html#generate/4-options) instead.

FranekKubis: I did:
```
  @github "https://github.com/software-mansion/popcorn"
  @github_pattern "https://github.com/software-mansion/popcorn/tree/main/local-live-view/%{path}#L%{line}"
```
for:
```
source_url: @github,
source_url_pattern: @github_pattern,
```
and sources links are now: `https://github.com/software-mansion/popcorn/blob/llv-v0.1.0-rc0`

FranekKubis: NVM - the pattern does not work like that - im using `source_ref: "llv-v#{version}/local-live-view"`.

mat-hek (approved):

PR #755, opened 2026-08-12
[LLV] Documentation - modules audit

FranekKubis:
- Added missing modules documentation. 
- Fixed wrong documentation. 
- Added Popcorn code part to documentation.

### DIFF
--- a/local-live-view/lib/local_live_view/local_live_view.ex
+++ b/local-live-view/lib/local_live_view/local_live_view.ex
@@ -290,8 +290,9 @@ defmodule LocalLiveView do
   @doc """
   Invoked once when the view is initialized, before the first `c:render/1`.
 
-  Use it to set up the initial assigns. Assigns passed down from the host
-  LiveView are not available here — they arrive through `c:update/2`.
+  Use it to set up the initial assigns. Assigns coming from the host LiveView
+  are not delivered here — `c:update/2` runs with them right after this
+  callback, before the first render.
 
   ```
   def mount(_params, _session, socket) do
@@ -306,12 +307,52 @@ defmodule LocalLiveView do
             ) ::
               {:ok, Socket.t()} | {:ok, Socket.t(), keyword()}
 
+  @doc """
+  Invoked with the query params of the page the view is rendered on.
+
+  Called after `c:mount/3` and again after every `push_patch/2`. `params` holds
+  the query string decoded into a map with string keys and `uri` is the full
+  URL. A local view is not mounted at the router, so this is the only callback
+  that sees the address bar.
+
+  ```
+  def handle_params(%{"tab" => tab}, _uri, socket) do
+    {:noreply, assign(socket, :tab, tab)}
+  end
+  ```
+  """
   @callback handle_params(
               params :: unsigned_params(),
               uri :: String.t(),
               socket :: Socket.t()
             ) :: {:noreply, Socket.t()}
 
+  @doc ~S'''
+  Receives the assigns the host LiveView passes down.
+
+  Every attribute other than `view` given to `local_live_view/1` is forwarded
+  here, the same way `Phoenix.LiveComponent` receives its assigns:
+
+  ```
+  <.local_live_view view="Cart" items={@items} currency="EUR" />
+  ```
+
+  It runs right after `c:mount/3` with the initial values, and then on every
+  re-render of the host LiveView. The host always sends the full set of
+  forwarded assigns, not a diff, so a view that needs to react only to actual
+  changes has to compare against its own assigns.
+
+  ```
+  def update(%{items: items}, socket) do
+    {:ok, assign(socket, :items, items)}
+  end
+  ```
+
+  The default implementation assigns everything it receives. These assigns are
+  also the authoritative server state that `c:handle_push_error/4` falls back
+  to, so overriding this callback with a guard on unchanged data means the
+  rollback needs handling explicitly — see `c:handle_push_error/4`.
+  '''
   @callback update(assigns :: map(), socket :: Socket.t()) :: {:ok, Socket.t()}
 
   @doc """
@@ -333,6 +374,28 @@ defmodule LocalLiveView do
   @callback handle_event(event :: binary, unsigned_params(), socket :: Socket.t()) ::
               {:noreply, Socket.t()} | {:reply, map, Socket.t()}
 
+  @doc """
+  Handles a message sent to the view's process.
+
+  A local live view runs as its own Elixir process inside the browser, so
+  anything that can reach that process arrives here — most often a timer set
+  with `Process.send_after/3`:
+
+  ```
+  def mount(_params, _session, socket) do
+    Process.send_after(self(), :tick, 1000)
+    {:ok, assign(socket, :time, Time.utc_now())}
+  end
+
+  def handle_info(:tick, socket) do
+    Process.send_after(self(), :tick, 1000)
+    {:noreply, assign(socket, :time, Time.utc_now())}
+  end
+  ```
+
+  The resulting render is pushed to the DOM as a diff, exactly as after
+  `c:handle_event/3`. The default implementation ignores the message.
+  """
   @callback handle_info(message :: term, socket :: Socket.t()) :: {:noreply, Socket.t()}
 
   @doc """
--- a/local-live-view/lib/local_live_view/local_live_view.ex
+++ b/local-live-view/lib/local_live_view/local_live_view.ex
@@ -10,8 +10,8 @@ defmodule LocalLiveView do
 
   `LocalLiveView` API similar to `Phoenix.LiveView`:
   - it runs in a separate Elixir process,
-  - it has `c:mount/3` and `c:render/1` callbacks, which behave
-    the same way as in a regular live view,
+  - it has `c:mount/3`, `c:handle_params/3` and `c:render/1` callbacks,
+    which behave the same way as in a regular live view,
   - it can spawn regular `Phoenix.Component`s and `Phoenix.LiveComponent`s,
   - events from these components by default go to their parent
     local live view.
@@ -312,8 +312,7 @@ defmodule LocalLiveView do
 
   Called after `c:mount/3` and again after every `push_patch/2`. `params` holds
   the query string decoded into a map with string keys and `uri` is the full
-  URL. A local view is not mounted at the router, so this is the only callback
-  that sees the address bar.
+  URL.
 
   ```
   def handle_params(%{"tab" => tab}, _uri, socket) do
@@ -329,6 +328,9 @@ defmodule LocalLiveView do
 
   @doc ~S'''
   Receives the assigns the host LiveView passes down.
+
+  The mechanism is the same as in `Phoenix.LiveComponent` and its
+  `c:Phoenix.LiveComponent.update/2` callback.
 
   Every attribute other than `view` given to `local_live_view/1` is forwarded
   here, the same way `Phoenix.LiveComponent` receives its assigns:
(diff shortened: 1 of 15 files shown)


local-live-view/lib/local_live_view/local_live_view.ex
```diff
@@ -306,12 +307,52 @@ defmodule LocalLiveView do
...
+  here, the same way `Phoenix.LiveComponent` receives its assigns:
+
+  ```
+  <.local_live_view view="Cart" items={@items} currency="EUR" />
+  ```
+
+  It runs right after `c:mount/3` with the initial values, and then on every
+  re-render of the host LiveView. The host always sends the full set of
+  forwarded assigns, not a diff, so a view that needs to react only to actual
+  changes has to compare against its own assigns.
```
mat-hek, line 345:
```suggestion
  re-render of the host LiveView.
```

What the host sends is an implementation detail, and it as works the same as LiveComponent, I don't think it requires further explanation

local-live-view/lib/local_live_view/local_live_view.ex
```diff
@@ -306,12 +307,52 @@ defmodule LocalLiveView do
...
+  """
   @callback handle_params(
               params :: unsigned_params(),
               uri :: String.t(),
               socket :: Socket.t()
             ) :: {:noreply, Socket.t()}
 
+  @doc ~S'''
+  Receives the assigns the host LiveView passes down.
+
```
mat-hek, line 331:
```suggestion
  The mechanism is the same as in `Phoenix.LiveComponent` and its `c:Phoenix.LiveComponent.update/2` callback.
```

local-live-view/lib/local_live_view/local_live_view.ex
```diff
@@ -306,12 +307,52 @@ defmodule LocalLiveView do
...
+  ```
+  def update(%{items: items}, socket) do
+    {:ok, assign(socket, :items, items)}
+  end
+  ```
+
+  The default implementation assigns everything it receives. These assigns are
+  also the authoritative server state that `c:handle_push_error/4` falls back
+  to, so overriding this callback with a guard on unchanged data means the
+  rollback needs handling explicitly — see `c:handle_push_error/4`.
```
mat-hek:
```suggestion
  The default implementation assigns everything it receives.
  
  This callback is also called by the default implementation of `c:handle_push_error/4`.
```

local-live-view/lib/local_live_view/local_live_view.ex
```diff
@@ -333,6 +374,28 @@ defmodule LocalLiveView do
   @callback handle_event(event :: binary, unsigned_params(), socket :: Socket.t()) ::
               {:noreply, Socket.t()} | {:reply, map, Socket.t()}
 
+  @doc """
+  Handles a message sent to the view's process.
+
+  A local live view runs as its own Elixir process inside the browser, so
+  anything that can reach that process arrives here — most often a timer set
```
mat-hek:
```suggestion
  anything that can reach that process arrives here — for example, a timer set
```

local-live-view/lib/server/channel.ex
```diff
@@ -1,20 +1,64 @@
...
+  # channel and nothing in the JS bundle subscribes to it, so the frame is
+  # dropped. Hidden from the docs until the browser side handles it and feeds
+  # the assigns into the view's update/2.
+  @doc false
   def set_mirror_assigns(mirror_id, assigns) do
     case Registry.lookup(LocalLiveView.ChannelRegistry, mirror_id) do
       [{pid, _}] -> GenServer.call(pid, {:set_mirror_assigns, assigns})
       [] -> {:error, :not_found}
     end
   end
```
mat-hek: this seems like a leftover, should be removed if it doesn't work

local-live-view/lib/server/channel.ex
```diff
@@ -1,20 +1,64 @@
...
+  Each LocalLiveView that syncs its assigns joins this channel on its own
+  topic, identified by the `mirror_id` obtained from
+  `LocalLiveView.Component.mirror_id/2`. The channel keeps the mirror assigns
+  returned by `c:LocalLiveView.Mirror.handle_sync/3` and makes them readable
+  from the server with `get_mirror_assigns/1`.
+
+  The channel is mounted by `LocalLiveView.Socket`, so there is nothing to wire
+  up by hand — `mix llv.install` adds the socket to your endpoint and the
+  `LocalLiveView.ChannelRegistry` to your supervision tree.
+  """
```
mat-hek: I don't think the user needs to know about this module's existence. The only relevant function is `get_mirror_assigns`, but perhaps it can be moved to `local_live_view.ex`, or a separate `LocalLiveView.Mirror`?

local-live-view/lib/server/controller.ex
```diff
@@ -1,14 +1,12 @@
...
-  and passes it to the template renderer.
-  """
+  # Renders a LocalLiveView mount point for a route declared with
+  # `LocalLiveView.Router.live_local/2`. The view module name is carried in
+  # `conn.private.llv_view` and handed to the template renderer.
+  @moduledoc false
 
   use Phoenix.Controller, formats: [html: "ControllerHTML"]
 
+  @doc false
```
mat-hek: It's already `@moduledoc false`

```suggestion
```

local-live-view/lib/server/controller_html.ex
```diff
@@ -1,15 +1,12 @@
...
-  the LocalLiveView mount point.
-  """
+  # Template renderer for `LocalLiveView.Controller`.
+  @moduledoc false
 
   use Phoenix.Component
 
   import LocalLiveView.Component
 
+  @doc false
```
mat-hek:
```suggestion
```

local-live-view/lib/server/router.ex
```diff
@@ -15,6 +15,50 @@ defmodule LocalLiveView.Router do
...
+
+        # Receives what the local view sends with push_server_event/3
+        def handle_event("archive", %{"id" => id}, socket) do
+          {:noreply, assign(socket, :items, Store.archive(id))}
+        end
+      end
+
+  `view_module` is taken by its last alias segment, so `HelloLocal` and
+  `MyApp.HelloLocal` both resolve to the `"HelloLocal"` view in the `local/`
+  project. A string or atom works too.
```
mat-hek, line 60: This seems very weird, is it how it works for regular LiveViews?

local-live-view/lib/local_live_view/local_live_view.ex
```diff
@@ -306,12 +307,52 @@ defmodule LocalLiveView do
...
+  Receives the assigns the host LiveView passes down.
+
+  Every attribute other than `view` given to `local_live_view/1` is forwarded
+  here, the same way `Phoenix.LiveComponent` receives its assigns:
+
+  ```
+  <.local_live_view view="Cart" items={@items} currency="EUR" />
+  ```
+
+  It runs right after `c:mount/3` with the initial values, and then on every
```
mat-hek, line 342: Isn't handle_params run before?

local-live-view/lib/local_live_view/local_live_view.ex
```diff
@@ -306,12 +307,52 @@ defmodule LocalLiveView do
             ) ::
               {:ok, Socket.t()} | {:ok, Socket.t(), keyword()}
 
+  @doc """
+  Invoked with the query params of the page the view is rendered on.
+
+  Called after `c:mount/3` and again after every `push_patch/2`. `params` holds
+  the query string decoded into a map with string keys and `uri` is the full
+  URL. A local view is not mounted at the router, so this is the only callback
+  that sees the address bar.
```
mat-hek: This effectively means that `mount` doesn't receive params? It should probably mention in the docs for `mount`, not here

local-live-view/lib/local_live_view/local_live_view.ex
mat-hek:
```suggestion
 - it has `c:mount/3`, `c:handle_params/3` and `c:render/1` callbacks,
   which behave the same way as in a regular live view,
```

local-live-view/mix.exs
```diff
@@ -64,6 +64,14 @@ defmodule LocalLiveView.MixProject do
       groups_for_extras: [
         Introduction: ~r"/introduction/",
         Guides: ~r"/guides/"
+      ],
+      groups_for_modules: [
+        "Browser / WASM (Popcorn)": [
+          ~r/^LocalLiveView$/
```
mat-hek, line 72: For some reason, I only see the server live views when generating docs:

Image

mat-hek (changes requested):

local-live-view/mix.exs
FranekKubis: You need to generate docs with MIX_TARGET=docs.

local-live-view/lib/local_live_view/local_live_view.ex
FranekKubis: I'd rather leave it as it is because it is an implementation detail but this _implementation_ is done by user in their own defined llvs.

local-live-view/mix.exs
mat-hek, line 11:
```suggestion
      aliases: aliases(),
      preferred_cli_target: [docs: :docs],
```

local-live-view/mix.exs
```diff
@@ -31,8 +31,8 @@ defmodule LocalLiveView.MixProject do
   # Run "mix help deps" to learn about dependencies.
   defp deps do
     [
-      {:popcorn, "~> 0.3.2", targets: :wasm},
-      {:ex_doc, "~> 0.34", only: [:dev, :test], runtime: false, warn_if_outdated: true},
+      {:popcorn, "~> 0.3.2", targets: [:wasm, :docs]},
+      {:ex_doc, "~> 0.34", only: [:docs, :dev, :test], runtime: false, warn_if_outdated: true},
```
mat-hek:
```suggestion
      {:ex_doc, "~> 0.34", only: [:dev, :test], targets: [:docs], runtime: false, warn_if_outdated: true},
```

local-live-view/lib/server/channel.ex
FranekKubis: Ok, get_mirror_assings moved to Mirror module. Channel gets `@moduledoc false`.

local-live-view/lib/server/router.ex
FranekKubis: It was done by marceli and I must've missed it.
It needs correction in code, but lets leave it for another PR.

local-live-view/lib/local_live_view/local_live_view.ex
FranekKubis: 1. mount
2. update
3. handle_params

local-live-view/lib/local_live_view/local_live_view.ex
mat-hek: By implementation detail I meant this:

> The host always sends the full set of forwarded assigns, not a diff

local-live-view/lib/local_live_view/local_live_view.ex
FranekKubis: That was some leftover. A mount function gets the same params as handle_params and llv can be mounted at router.
Removed: `A local view is not mounted at the router, so this is the only callback that sees the address bar.`

mat-hek (approved):

local-live-view/lib/server/socket.ex
```diff
@@ -1,22 +1,22 @@
 defmodule LocalLiveView.Socket do
-  @moduledoc """
-  The `Phoenix.Socket` local live views connect to.
+  @moduledoc false
```
mat-hek: I think this one should be public, since user/generator puts it in the user code, it just shouldn't reference `LocalLiveView.Channel` or explain its internals in the docs

PR #756, opened 2026-08-12
[LLV] Make kanban usable as a deployed demo

mat-hek:
- Only display visited boards, kept in local storage
- Add 'create sample board' button
- Allow changing board names


jgonet (approved):
PR title seems totally unrelated to contents.

PR #764, opened 2026-08-14
[LLV] release blog post

mat-hek:
This is not to be merged, but I don't have access to SWM blog yet, so here it is for a review.

The content is 100% protein, agent only reviewed. There are no links yet, I'll add them once LLV and demos are published.

TODO:
- [x] in example, mention you need to install LLV 
- [x] mention SSR
~~mention beam migration~~
- [x] add links

UPDATE:

I reshaped the post:
- mentioned what's LLV on the start
- rewrote the 'why LLV?' section, it's now hopefully more to the point
- changed the example to show that live components can be used without changes
- decided not to mention beam migration - it's more about popcorn than LLV and there's a lot in the post already

### DIFF
--- a/llv-blog.md
+++ b/llv-blog.md
@@ -1,14 +1,16 @@
 # The first release of Local LiveView!
 
-This April, we announced Local LiveView: a library to run LiveView code right in the browser. It was a POC at the time, and since then, we've put lots of work into making it usable in the wild. Today, I'm happy to share that the first Hex release of Local LiveView just landed 🎉 It ships with a solid core of features, documentation, guides, and even an Igniter-based installer. We've deployed some demos, too. So, it's the perfect time to start hacking around.
+Local LiveView is a library to run LiveView code in the browser. We announced it back in April, when it was still a POC. Since then, we've put lots of work into making it usable in the wild. Today, I'm happy to share that the first Hex release of Local LiveView just landed 🎉 It ships with a solid core of features, documentation, guides, and even an Igniter-based installer. We've deployed some demos, too. So, it's the perfect time to start hacking around.
 
 ## Why Local LiveView?
 
-LiveView is great for cases where you're fine with keeping pretty much all the state on the server, with all its benefits and tradeoffs. When client-side state is needed, things get complex. Even though there are JS hooks and JS commands, they're far from idiomatic LiveView code, and therefore harder to reason about and maintain. And you're often forced to write JS.
+LiveView is great for cases where you're fine with keeping pretty much all the state on the server, which has its benefits and tradeoffs. When client-side state is needed, things get complex. Even though there are JS hooks and JS commands, they're far from idiomatic LiveView code, and therefore harder to reason about and maintain. And you're often forced to write JS.
 
 Thus, Local LiveView allows you to run LiveView code in the browser via Popcorn, keeping all assigns locally. You can now freely choose what should stay on the client, and what needs to reach the server. It helps, for example, offload the server from handling simple UI updates, drastically reduce latency on poor networks, and avoid 'WebSocket disconnected' issues.
 
-## How do I use it?
+## Moving a live component to the browser, step by step
+
+### 1. Take a live component
 
 An example is worth a thousand words, so let's follow one. Assuming you have Local LiveView installed in your project (did I mention there's an Igniter installer?), consider a simple live component:
 
@@ -38,22 +40,9 @@ defmodule MyAppWeb.ThermostatComponent do
 end
 ```
 
-rendered from a LiveView:
+### 2. Create a Local LiveView
 
-```elixir
-defmodule MyAppWeb.MyLive do
-  use Phoenix.LiveView
-
-  @impl true
-  def render(assigns) do
-    ~H"""
-    <.live_component module={MyAppWeb.ThermostatComponent} id="thermostat" temperature={25} />
-    """
-  end
-end
-```
-
-Let's make the component run on the client! For that, we need a simple Local LiveView that spawns our component:
+Let's make the component run on the client! For that, we need a simple Local LiveView that renders our component:
 
 ```elixir
 defmodule MyAppWeb.MyLocal do
@@ -68,7 +57,9 @@ defmodule MyAppWeb.MyLocal do
 end
 ```
 
-Now, let's change our parent LiveView. Instead of rendering the Thermostat directly, let it render the new local view:
+### 3. Render the Local LiveView
+
+Now, let's render our local view. We can do it from a regular LiveView:
 
 ```elixir
 defmodule MyAppWeb.MyLive do
@@ -83,11 +74,15 @@ defmodule MyAppWeb.MyLive do
 end
 ```
 
+### 4. Put the live component in the client part of the project
+
 The only step left is to move the Thermostat to the right place (`local/lib/*`) in your project. Local LiveView enforces separation of the client and server code, so you don't leak the server stuff to the client accidentally, but can still reuse the client code on the server.
+
+### 5. It works!
 
 And... that's it! The Thermostat itself is unchanged, but it now runs locally! Let's see how it works.
 
-The local view we just created is an entry point to the client-side world. In this example, it doesn't do much: just gets the `temperature` assign from the server and uses it to render the Thermostat component. However, local views can have complex logic, handle events, and render multiple live and regular components. In fact, we could merge the Thermostat component into our local view - it's quite straightforward, so we leave it as an exercise for the reader ;) The docs explain the Local LiveView API and its relation to LiveView and LiveComponent in detail.
+The local view we just created is an entry point to the client-side world. In this example, it doesn't do much: just gets the `temperature` assign from the server and uses it to render the Thermostat component. However, local views can have complex logic, handle events, and render multiple live and regular components. You could even move the Thermostat's logic into the local view, merging them together - it can be a nice excercise to get started ;) The docs explain the Local LiveView API and its relation to LiveView and LiveComponent in detail.
 
 It's worth noting that a server-side LiveView can render many local views, and a local view can reach back to the server - there are two mechanisms for that:
 - `push_server_event` - sends events to the server, which replies with updated assigns (LiveVue/LiveSvelte style),
@@ -105,9 +100,9 @@ This post shows a very simple example, but Local LiveView is already capable of 
 
 - Burrito order form - a fairly complex form demo, comparing regular LiveView and Local LiveView side-by-side. It also demonstrates synchronizing state via mirror sync.
 
-## The future
+## Coming next: real-world validation, smaller bundles, SSR
 
-Even though Local LiveView has been pretty stable for us recently, we're going to test it further in different scenarios and see how it behaves. We're going to polish the API along the way, but with no big changes expected.
+Even though Local LiveView feels pretty stable already, we're going to test it further in different scenarios and see how it behaves. We're going to polish the API along the way, but with no big changes expected.
 
 Removing any dependencies on LiveView private APIs is another important point. Fortunately, we're well on our way to doing that and we're working closely with the LiveView team.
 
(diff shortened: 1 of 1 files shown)


llv-blog.md
```diff
@@ -0,0 +1,103 @@
+# The first release of Local Live View!
+
+Today marks another milestone in the Local Live View development. In April, Franek Kubis announced it at ElixirConf EU, and we published several demos and POCs since then. Now, it finally has docs, reasonable API (more or less :P) and a Hex release. It's not been run in prod yet, but we're close already. We deployed some demos, too. So, it's the perfect time to start hacking around.
+
```
jgonet, line 4: Paragraph on what is local live view, what problem does it solve?

llv-blog.md
```diff
@@ -0,0 +1,103 @@
...
+## Why Local Live View?
+
+Local Live View makes it possible to manage local state the same way as you do with the server state in a regular Live View. While Live View provides JS hooks and JS commands, they're arguably more complex and harder to maintain than regular Elixir code. And you're often forced to write JS.
```
jgonet: > While Live View provides JS hooks and JS commands, they're arguably more complex and harder to maintain than regular Elixir code. And you're often forced to write JS.

Doesn't follow well from prev sentence for me

llv-blog.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+Now, moving the component to the client requires only slight changes. First, we `use LocalLiveView`, and change the module name to reflect that:
+
+```diff
+- defmodule MyAppWeb.ThermostatComponent do
+-   use Phoenix.LiveComponent
++ defmodule MyAppWeb.ThermostatLocal do
++   use LocalLiveView
+```
```
jgonet: Not said if module name change is necessary, maybe keep it in text at the end for simplicity?

llv-blog.md
```diff
@@ -0,0 +1,103 @@
...
+
+Then, we need to remove `phx-target={@myself}` for the `click` event - Local Live View events always target themselves, not their parents:
```
jgonet: Q: does plain LV defaults to itself? I wonder if we should enforce having this even if it has only one value – would enable copy & paste between LLV and LV.

llv-blog.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+Thermostat can now be rendered with:
+
+```diff
+- <.live_component module={MyAppWeb.ThermostatComponent} id="thermostat" temperature={25} />
++ <.local_live_view view="MyAppWeb.ThermostatLocal" id="thermostat" temperature={25} />
+```
+
+The only step left is to move the Thermostat to the right place in your project, and voila! The Thermostat is now fully local.
```
jgonet: voila -> voilà

llv-blog.md
```diff
@@ -0,0 +1,103 @@
...
+
+Notice that we changed the Thermostat from a Live Component to a Local Live View. While the API is similar, they're different under the hood: Local Live View runs in a separate process (and even on a separate VM). All the differences are outlined in the docs, the most important being the communication, which now goes across the network. For that, you can use one of two mechanisms: push_server_event or mirror sync - they're thoroughly explained in the docs as well.
```
jgonet: "For that, you can use one of two mechanisms" – maybe single word explanation in parens?

llv-blog.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Burrito order form - a fairly complex form demo, comparing regular Live View and Local Live View side-by-side. It also demonstrates synchronizing state via mirror sync.
+
+## The future
+
+Even though Local Live View has been pretty stable for us recently, we're going to test it further in different scenarios and see how it behaves. We're going to polish the API along the way, but with no big changes expected.
+
+Removing any dependencies on Live View private APIs is another important point. Fortunately, we're well on our way to doing that and we're working closely with the Live View team.
+
+Reducing the bundle size is something we constantly keep working on. We recently introduced an experimental Elixir tree-shaking tool that already helps significantly, with the Kanban demo's size down to < 1.5 MB compressed.
```
darthez: Nit: maybe worth adding what was the size before the change or what was the % decrease in size

llv-blog.md
mat-hek: The next paragraph covers it. You mean I should start with that or there's something missing there?

llv-blog.md
mat-hek: > Q: does plain LV defaults to itself?

Yes

> I wonder if we should enforce having this

Since LLV is a Live View underneath, it may be hard to do without using LV private API. We can see after the release

llv-blog.md
FranekKubis: I think changing the module name makes sense, as it also highlights that LLV is not just a component (it has a process under the hood).

llv-blog.md
FranekKubis: This paragraph seems oddly specific. Do we need to explain how phx-target works in the first blog post - seems like a detail that makes explaining LLV mechanism more complicated than it is.

llv-blog.md
```diff
@@ -0,0 +1,103 @@
+# The first release of Local Live View!
+
+Today marks another milestone in the Local Live View development. In April, Franek Kubis announced it at ElixirConf EU, and we published several demos and POCs since then. Now, it finally has docs, reasonable API (more or less :P) and a Hex release. It's not been run in prod yet, but we're close already. We deployed some demos, too. So, it's the perfect time to start hacking around.
+
+## Why Local Live View?
+
+Local Live View makes it possible to manage local state the same way as you do with the server state in a regular Live View. While Live View provides JS hooks and JS commands, they're arguably more complex and harder to maintain than regular Elixir code. And you're often forced to write JS.
+
+Thus, Local Live View allows you to run Live Views in the browser via Popcorn, keeping all their assigns locally. You can now freely choose what should stay on the client, and what needs to reach the server. It helps, for example, offload the server from handling simple UI updates, drastically reduce latency on poor networks, and avoid 'WebSocket disconnected' issues.
```
FranekKubis: I would add a short paragraph about code separation between browser and server parts (`local` directory) in the same phoenix repo.

FranekKubis (commented):
Great post! Really like the length.

Noticed the TODO bullet points — hopefully they stay similarly concise

llv-blog.md
mat-hek: Right, I wouldn't just place the diff without explanations. I'm open to suggestions on how to reword it ;)

llv-blog.md
mat-hek: Thought about it, but I'm not sure where to put it. I'll think again :P

llv-blog.md
FranekKubis: Maybe something like this:
```suggestion
We should also drop phx-target={@myself} — since it's no longer a LiveComponent, Local Live Views target themselves by default.
```
I think this is succinct enough.

llv-blog.md
```diff
@@ -53,51 +53,66 @@ defmodule MyAppWeb.MyLive do
...
-Thermostat can now be rendered with:
+The only step left is to move the Thermostat to the right place in your project. Local LiveView enforces separation of the client and server code, so you don't leak the server stuff to the client accidentally, but can still reuse the client code on the server.
 
-```diff
-- <.live_component module={MyAppWeb.ThermostatComponent} id="thermostat" temperature={25} />
-+ <.local_live_view view="MyAppWeb.ThermostatLocal" id="thermostat" temperature={25} />
-```
+And... that's it! The Thermostat itself is unchanged, but it now runs locally! Let's see how it works.
+
+The local view we just created is an entry point to the client-side world. In this example, it doesn't do much: just gets the `temperature` assign from the server and uses it to render the Thermostat component. However, local views can have complex logic, handle events, and render multiple live and regular components. In fact, we could merge the Thermostat component into our local view - it's quite straightforward, so we leave it as an exercise for the reader ;) The docs explain the LocalLiveView API and its relation to LiveView and LiveComponent in detail.
```
FranekKubis: > In fact, we could merge the Thermostat component into our local view - it's quite straightforward, so we leave it as an exercise for the reader ;)

Sounds a bit strange – maybe we can just remove this sentence?

llv-blog.md
```diff
@@ -0,0 +1,120 @@
+# The first release of Local LiveView!
+
+This April, we announced Local LiveView: a library to run LiveView code right in the browser. It was a POC at the time, and since then, we've put lots of work into making it usable in the wild. Today, I'm happy to share that the first Hex release of Local LiveView just landed 🎉 It ships with a solid core of features, documentation, guides, and even an Igniter-based installer. We've deployed some demos, too. So, it's the perfect time to start hacking around.
```
FranekKubis:
```suggestion
This April, we announced Local LiveView: a library to run LiveView code in the browser. It was a POC at the time, and since then, we've put lots of work into making it usable in the wild. Today, I'm happy to share that the first Hex release of Local LiveView just landed 🎉 It ships with a solid core of features, documentation, guides, and even an Igniter-based installer. We've deployed some demos, too. So, it's the perfect time to start hacking around.
```

llv-blog.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Why Local LiveView?
+
+LiveView is great for cases where you're fine with keeping pretty much all the state on the server, with all its benefits and tradeoffs. When client-side state is needed, things get complex. Even though there are JS hooks and JS commands, they're far from idiomatic LiveView code, and therefore harder to reason about and maintain. And you're often forced to write JS.
```
FranekKubis:
```suggestion
LiveView is great for cases where you're fine with keeping pretty much all the state on the server, which has its benefits and tradeoffs. When client-side state is needed, things get complex. Even though there are JS hooks and JS commands, they're far from idiomatic LiveView code, and therefore harder to reason about and maintain. And you're often forced to write JS.
```

llv-blog.md
```diff
@@ -0,0 +1,120 @@
...
+
+Thus, Local LiveView allows you to run LiveView code in the browser via Popcorn, keeping all assigns locally. You can now freely choose what should stay on the client, and what needs to reach the server. It helps, for example, offload the server from handling simple UI updates, drastically reduce latency on poor networks, and avoid 'WebSocket disconnected' issues.
+
+## How do I use it?
+
+An example is worth a thousand words, so let's follow one. Assuming you have Local LiveView installed in your project (did I mention there's an Igniter installer?), consider a simple live component:
```
FranekKubis, line 15: While this example is really cool and flashy, I'm worried it might be a bit overwhelming at first glance.

Maybe we should revert to the LiveComponent rewrite and accept the tradeoff with the phx-target explanation?

Or maybe render `MyAppWeb.MyLocal` directly from router?

FranekKubis (changes requested):

llv-blog.md
mat-hek: The intent was to share that you usually don't need an intermediary, dumb LLV

llv-blog.md
mat-hek: The `phx-target` explanation was not the main problem, it was the impression that you can't reuse live components as they are. That's why I wouldn't get back to the original. Rendering prevents client -> server communication and I'd prefer showcasing what we expect people to usually do. But I can just remove the first LiveView snippet, it's not necessary

llv-blog.md
FranekKubis: > In fact, we could merge the Thermostat component into our local view as we usually wouldn't need an intermediary LLV.

what do you think?

llv-blog.md
FranekKubis: Ok I see it - for me the most complex thing was rendering `phoenixliveview -> llv -> live_component`, so thats why I proposed to render like this:
```elixir
defmodule MyApp2Web.Router do
  import LocalLiveView.Router

  scope "/", MyApp2Web do
    pipe_through :browser

    get "/", PageController, :home
    live_local "/hello_local", HelloLocal
    live "/hello_live", MyApp2Web.HelloLive
  end
end
```
but I agree it is good to show rendering from within Phoenix LiveView for the server connections sake.

> But I can just remove the first LiveView snippet, it's not necessary

IMO the snippet is playing its role in the explanation, to show how the thing were before, so it should stay.
So I guess we let the example stay as it is.

BTW: I will go back to think about router rendering and mirroring - maybe there is still some solution that makes it possible to render from router and also be connected to server in some way.

mat-hek: Thanks for the feedback, I sent the post to the marketing already, I think it's good enough :P

> BTW: I will go back to think about router rendering and mirroring - maybe there is still some solution that makes it possible to render from router and also be connected to server in some way.

I think it's fine to leave the router mounting for purely client-side LLVs for now. If you need to connect with the server, you probably have LiveView there to handle the server logic anyway. We'll hopefully get some feedback after the release too ;)

PR #768, opened 2026-08-21
Remove custom LLV server

mat-hek:
Elixir (Wasm part):
- Create views using `live_render`. (Ab)use sticky option, as it allows us to mount a view fairly easily and makes it fully controllable by its channel that we control.
- Remove `server.ex`. Instead, use the original LiveView channels and `proxy.ex` that implements LiveView behaviour and wraps each LLV. The dispatcher now acts as `Phoenix.Socket.Transport`.

JS:
- Because views are created with `live_render` on Elixir side, the view creation requires less hacks and less private APIs (in fact, it only depends on `socket.newRootView()` and `view.channel`)
- Extracted mirrors.ts and views.ts from index.ts
- Simplified PopcornClient
- Replaced `socket.view.pushEvent` (private) with `socket.js().push` (public)

As a result, we don't depend on any private APIs on Elixir side and very few on the JS side - only `socket.newRootView()` and `view.channel` are crucial for LLV in the current state. Assumptions on the undocumented LV behaviour are also very limited. I believe the code is now simpler too.

On the other hand, using more LV code instead of our hacks brings some unnecessary stuff, like signing tokens. Maybe we can convince the LV Team to make it optional

Two new stubs were necessary, they can be removed when we switch to the BEAM. Without the stubs, this PR removes more lines than it adds.

TODO:
- [x] re-enable tree-shaking
- [x] use public API of plug-crypto


local-live-view/assets/local_live_view/index.ts
```diff
@@ -52,122 +33,87 @@ class PopcornClient {
...
+        error: err instanceof Error ? err : new Error(String(err)),
+        durationMs: 0,
+      }))
+      .then((result) => {
+        if (!result.ok && !opts?.suppressErrorLog) {
+          console.error(`LLV ${message.action} error`, result.error);
+        }
+        return result;
+      });
   }
```
jgonet, line 61: Feels overcomplicated.
```
/**
 * Sends an action to the dispatcher and returns the dispatcher's reply.
 * Error replies are logged by default; pass suppressErrorLog to silence them.
 * Rejects if the runtime isn't ready or the call itself fails.
 */
async call(message: CallMessage, opts?: CallOptions): Promise<CallResult> {
  const { popcorn } = this;
  assertRuntimeReady(popcorn); // or even create a getter that asserts popcorn's ready.

  const result = await popcorn.call(message, {
    timeoutMs: DEFAULT_CALL_TIMEOUT_MS,
  });

  if (!result.ok && !opts?.suppressErrorLog) {
    console.error(`LLV ${message.action} error`, result.error);
  }

  return result;
}
```

local-live-view/assets/local_live_view/index.ts
```diff
@@ -52,122 +33,87 @@ class PopcornClient {
...
-    await engine.bootPopcorn();
+  /**
+   * Boots the WASM runtime and mounts every LocalLiveView on the page.
+   * Typically called right after `liveSocket.connect()`. Idempotent —
+   * repeated calls are no-ops.
+   */
+  connect(): void {
+    this.connectPromise ??= this.doConnect().catch((err: unknown) => {
+      console.error("LLV: connect failed", err);
+    });
```
jgonet: Why we're not using try-catch?

local-live-view/assets/local_live_view/index.ts
```diff
@@ -181,61 +127,27 @@ export class LLVEngine {
...
+        this.bufferedServerEvents.push(detail);
         return;
       }
-      sendServerMessage(this.pop, detail);
+      this.sendServerEvent(detail);
+    });
+  }
+
+  private sendServerEvent(detail: LLVServerEventDetail): void {
+    void this.pop.call({
```
jgonet: `void` unnecessary.

local-live-view/lib/local_live_view/application.ex
```diff
@@ -4,12 +4,41 @@ defmodule LocalLiveView.Application do
...
+      server: false,
+      # Only used for signing some JS <-> Wasm messages. LiveView doesn't support
+      # disabling the signing.
+      secret_key_base: "local-live-view-browser-only-secret-key-base-000000000000000000",
+      live_view: [signing_salt: "local-live-view", hibernate_after: :infinity]
+    ]
+  end
+
+  # FIXME: verify when the application startup fails, remove the manual table creation
+  defp ensure_plug_crypto_key_cache do
```
jgonet: Intended to be fixed after the PR?

local-live-view/lib/local_live_view/proxy.ex
```diff
@@ -0,0 +1,229 @@
...
+    end
+  end
+
+  @impl true
+  def handle_event(event, params, socket) do
+    view(socket).handle_event(event, params, socket)
+  end
+
+  @impl true
+  def handle_info({:llv, %{"action" => "update_assigns", "assigns" => encoded_assigns}}, socket) do
```
jgonet: Could be more readable with assigns match extracted inside the fn 🤷

local-live-view/lib/local_live_view/proxy.ex
```diff
@@ -0,0 +1,229 @@
...
+        #{inspect(other)}
+        """
+    end
+  end
+
+  defp resolve_view_module!(name) do
+    module = Module.concat([name])
+    loaded? = match?({:module, _module}, Code.ensure_loaded(module))
+
+    unless loaded? and function_exported?(module, :render, 1) do
```
jgonet: `unless` -> `if not`

local-live-view/test/mix/tasks/llv_install_test.exs
```diff
@@ -67,13 +67,16 @@ defmodule Mix.Tasks.Llv.InstallTest do
   end
 
   describe "inject_app_js" do
-    test "adds LLVEngine.create() import + call after liveSocket.connect()" do
+    test "runs LLVEngine.create() before liveSocket.connect() and llvEngine.connect() after" do
```
jgonet: "state initialized with correct order" with `LLVEngine.create() -> liveSocket.connect() -> llvEngine.connect()` comment?

jgonet (approved):
Mostly nits.

local-live-view/lib/local_live_view/application.ex
```diff
@@ -4,12 +4,41 @@ defmodule LocalLiveView.Application do
...
+    case Application.ensure_all_started(:plug_crypto) do
+      {:ok, _apps} ->
+        :ok
+
+      {:error, _reason} ->
```
FranekKubis: Maybe we could at least log the reason here.

local-live-view/lib/local_live_view/dispatcher.ex
```diff
@@ -37,128 +80,242 @@ defmodule LocalLiveView.Dispatcher do
...
+        }
+        """,
+        message
+      )
     end
+
+    :ok
+  end
+
+  defp validate_push(%{event: "redirect", topic: topic, payload: payload}) do
```
FranekKubis: If LLV is terminated here maybe we should remove its `state.views[id]` and unmount the view on the JS side.

local-live-view/lib/stubs/phoenix_config_stub.ex
```diff
@@ -0,0 +1,121 @@
+# Stub: AtomVM doesn't support :ets.select nor :ets.match_delete,
```
FranekKubis, line 1: stubs are copied from phoenix 1.8.4 - maybe we should change the phoenix version to `"~>1.8.4"` in `mix.exs` instead of `"~>1.8"`

local-live-view/lib/local_live_view/proxy.ex
```diff
@@ -0,0 +1,229 @@
...
+
+    case opts do
+      [] -> {:ok, socket}
+      opts -> {:ok, socket, opts}
+    end
+  end
+
+  @impl true
+  def handle_event(event, params, socket) do
+    view(socket).handle_event(event, params, socket)
```
FranekKubis, line 42: Shouldn't we check the reply as we do in line 65?

local-live-view/lib/local_live_view/dispatcher.ex
```diff
@@ -37,128 +80,242 @@ defmodule LocalLiveView.Dispatcher do
...
+    false
+  end
+
+  defp validate_push(_message) do
+    true
   end
 
-  defp parse_assigns(nil), do: %{}
-  defp parse_assigns(assigns), do: :erlang.binary_to_term(Base.decode64!(assigns))
+  defp topic_to_id("lv:" <> id), do: id
```
FranekKubis, line 308: Maybe create fallback clause here since we can have `"lvu:*"` topics.

local-live-view/lib/local_live_view/application.ex
mat-hek: I'll try to remove this altogether, still in TODO

local-live-view/lib/local_live_view/dispatcher.ex
mat-hek: I think we can just allow redirects from LLVs, WDYT?

local-live-view/lib/local_live_view/proxy.ex
mat-hek: LV handles it already. In line 65 we only do the handling to print the right callback in the error, here the callback is the same.

local-live-view/lib/local_live_view/application.ex
mat-hek: no, I'll try to fix in this PR

local-live-view/assets/local_live_view/index.ts
mat-hek: This shape is mostly to make call never reject, otherwise we'd be forced to handle both `!result.ok` and rejections everywhere. I'll try to simplify.

PR #769, opened 2026-08-21
llv: fix assigns null handling

jgonet:
#708 moved that to elixir, seems like it can be null.

### DIFF
--- a/local-live-view/assets/local_live_view/index.ts
+++ b/local-live-view/assets/local_live_view/index.ts
@@ -25,10 +25,12 @@ type CallResult = Awaited<ReturnType<Popcorn["call"]>>;
 
 interface CreateArgs {
   id: string;
-  view: string | null;
+  view: string;
   url: string;
   urlParams: Record<string, string>;
-  assigns: string;
+  // Can be null when mounting LLVs manually via <div data-pop-view="...">
+  // and not setting `data-pop-assigns`.
+  assigns: string | null;
   mirrorId: string | undefined;
 }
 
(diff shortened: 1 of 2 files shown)


FranekKubis (approved):

local-live-view/assets/local_live_view/index.ts
```diff
@@ -28,7 +28,7 @@ interface CreateArgs {
   view: string | null;
   url: string;
   urlParams: Record<string, string>;
-  assigns: string;
+  assigns: string | null;
```
mat-hek, line 33: From what I reckon, this is the only case when assigns are unset, correct me if I'm wrong

```suggestion
  // Assigns are not set when someone manually places <div data-pop-view="..."> on the page
  assigns: string | null;
```

local-live-view/assets/local_live_view/index.ts
```diff
@@ -28,7 +28,7 @@ interface CreateArgs {
   view: string | null;
```
mat-hek: `view` is always there, no matter what

```suggestion
  view: string;
```

local-live-view/assets/local_live_view/index.ts
```diff
@@ -192,7 +192,7 @@ export class LLVEngine {
       view: pop_view_el.getAttribute("data-pop-view"),
```
mat-hek:
```suggestion
      view: pop_view_el.getAttribute("data-pop-view")!,
```

local-live-view/assets/local_live_view/index.ts
jgonet: Yup – https://github.com/software-mansion/popcorn/blob/9d790eeff6fd9cfe8a8ac2075bc54dc320fb46af/landing-page/src/pages/demos/local-thermostat.astro#L26-L31

I think we should expand the API for uses like that (some API like `.mount` on `engine` or maybe a custom component. We should document it at least if that's the intended way of mounting LLVs from JS.

FranekKubis: Good idea, I created an issue for that:
https://github.com/software-mansion/popcorn/issues/773

